PR #318, opened 2014-03-20
Add Pebble notifications

HawkiesZA:
Added the ability for Telegram to send notifications to a connected Pebble smartwatch (using the [reference documentation](https://developer.getpebble.com/1/03_API_Reference/03_Android/) with slight modifications).

This can be toggled in Notification settings and defaults to "Off".

I have been testing this over the past few days with my Pebble connected to my Nexus 5/Nexus 7 (2012).


amuelli:
Why has this been removed again in version 3.1.1?

HawkiesZA:
If I had to wager a guess, I'd say it's because the Pebble app now supports notifications from all apps, so this feature is redundant.

PR #341, opened 2014-04-01
Refactoring, added some object-oriented logic

maxbruecken:
Telegram code is not pretty object-oriented yet. I'm trying to make the code more readable and cleaner


hostomsrd:
that work does not have many errors. android studio and in fact all and nothing.. help me please

AmirHosseinKarimi:
Sorry for send message here. I do not found issus in this repo.

Hi, I found many bug in Telegram new version.
Before i use Pluse messanger but for newer version install telegram official but found many problem and now back to pluse.
Problems:
1. Do not show message preview in notification bar(whene enabled)
2. Show unread messages even whene loged out
3. Telegram is hungry after install and eat my phone RAM absolutlly in my case
4. Do not show link info (preview) even in public chat

PR #360, opened 2014-04-10
add zh-rCN and zh-rTW translation for Telegram

panyingyun:
I am sorry for some file changed(such as build.gradle,gradle-wrapper.properties), plealse ignore them. 
only merge values-zh-rCN/strings.xml  and values-zh-rTW/strings.xml  will be OK.

### DIFF
--- a/TMessagesProj/build.gradle
+++ b/TMessagesProj/build.gradle
@@ -1,11 +1,4 @@
-buildscript {
-    repositories {
-        mavenCentral()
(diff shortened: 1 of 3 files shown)


suokunlong:
Hi, strings.xml for zh_CN is 100%. We waited many months but it's still not merged in the source codes. Could someone do this please? I pinged Markus Ra but no response. Thanks.

PR #551, opened 2014-08-25
Disable screenshots and better looking emoji keyboard

jackgu1988:
- Added option to disable the ability to take screenshots (Android 4.0+)
- Improved the looks of the emoji "keyboard"


TameOfGroans:
This is an arms race with root'd devices you cannot win. Converse with those you trust. 

Rely on secrecy [encryption] to prevent those not party to conversation from retaining secrets.

Issue #222, opened 2014-02-27
Add Google Hangouts emojis to Telegram


Gabryxx7:
Hi, I was wondering if it was possible to add Google hangouts Emojis to telegram... It would be cool to be able to choose which set of emojis to use (for example by selecting it in the settings)... I tried to modify sprites but I couldn't find good resolution hangouts emojis...

sabret00the:
Just to point out, as of January 5th, 65% of devices support native emoji?

Multimoon:
A toggle to use system emojjis versus the packaged ones makes complete sense. It'd be a simple check of BUILD_VERSION_CODE, to display a optional check box titled "Use system emoji". This would not increase the package size at all. Would not interfere with old device support. And would satisfy everyone in using the emojii set they prefer. There is also no reason to state that this is a " overhead", as the option to change the chat background is also presented. Nor is a single check box much of a complex overhead.

wp9015362:
The Emojis should be consistent across all platforms, all devices and all clients IMHO.

I don't want to ask myself if a conversation partner sees the same Emoji as I do when I send one to him. I want to be sure that when I send him an Emoji, that he sees it the way I see it and the way it was intended.

Therefore I am against an option for changing the Emojis.

I think Telegram is doing it the right way currently. Please don't change it.

sabret00the:
@wp9015362 I guess the question I ask in that case is why should the default be the iOS emoji?

Multimoon:
Leaving iOS as default is fine, but the user should be presented with a [b]choice[/b] over using the native ones. @wp9015362 in no way does this interfere with the 'consistency' of the app.

sabret00the:
@Multimoon If there are more Telegram users on Android than iOS, would it not make sense to have the Android emoji as the default?

Multimoon:
The default set really doesn't matter, what matters is allowing the users
to make their own decision on emojis, as they dramatically influence the
apps looks.

wp9015362:
@sabret00the Because the iOS Emojis look great, are well known and people are used to them (from WhatsApp and so on).

Also, as DrKLO mentioned earlier, the iOS Emoji set apparently consists of 845 different Emojis and there currently is no other Emoji set that offers replacements for all of those 845 Emojis. At least that's what I understood from DrKLO's post (https://github.com/DrKLO/Telegram/issues/222#issuecomment-50251477).

[...]

PR #625, opened 2014-10-20
Stock emoji set

robin0van0der0v:
Added an option to use the stock emoji set.
#624

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/android/Emoji.java
+++ b/TMessagesProj/src/main/java/org/telegram/android/Emoji.java
@@ -13,6 +13,7 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Locale;
 
+import android.content.SharedPreferences;
 import android.graphics.BitmapFactory;
 import android.graphics.Canvas;
 import android.graphics.ColorFilter;
@@ -31,13 +32,14 @@ import org.telegram.messenger.Utilities;
 import org.telegram.ui.ApplicationLoader;
 
 public class Emoji {
-	private static HashMap<Long, DrawableInfo> rects = new HashMap<Long, DrawableInfo>();
-	private static int drawImgSize, bigImgSize;
-	private static boolean inited = false;
-	private static Paint placeholderPaint;
-	private static EmojiBitmap emojiBmp[] = new EmojiBitmap[5];
-	private static boolean loadingEmoji[] = new boolean[5];
+    private static HashMap<Long, DrawableInfo> rects = new HashMap<Long, DrawableInfo>();
+    private static int drawImgSize, bigImgSize;
+    private static boolean inited = false;
+    private static Paint placeholderPaint;
+    private static EmojiBitmap emojiBmp[] = new EmojiBitmap[5];
+    private static boolean loadingEmoji[] = new boolean[5];
     private static int emojiFullSize;
+    private boolean systemEmoji = false;
 
     private static class EmojiBitmap {
         public int[] colors;
@@ -202,8 +204,8 @@ public class Emoji {
                     0x00000000D83DDD31L, 0x00000000D83DDD32L, 0x00000000D83DDD33L, 0x00000000000025FCL, 0x00000000000025FBL, 0x00000000000025FEL, 0x00000000000025FDL,
                     0x00000000000025AAL, 0x00000000000025ABL, 0x00000000D83DDD3AL, 0x0000000000002B1CL, 0x0000000000002B1BL, 0x00000000000026ABL, 0x00000000000026AAL,
                     0x00000000D83DDD34L, 0x00000000D83DDD35L, 0x00000000D83DDD3BL, 0x00000000D83DDD36L, 0x00000000D83DDD37L, 0x00000000D83DDD38L, 0x00000000D83DDD39L}};
-	
-	static {
+    
+    static {
         if (AndroidUtilities.density <= 1.0f) {
             emojiFullSize = 30;
         } else if (AndroidUtilities.density <= 1.5f) {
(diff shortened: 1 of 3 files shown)


Multimoon:
I'm baffled why this is such a debate. There is no right or wrong over which set is better. Simply leave the iOS pack in the app. And implement a setting to use stock system emojiis. I assume you've heard of the GNOME project? with the release of 3.0 years ago, they attempted to force their opinion on users. It didn't go well.

DrFiz:
As an Android user and Telegram lover I'd like to see merged this pull request

I honestly like more Android emojis, but... what if everybody could use the preferred emoji set and the receiver actually see the emote as sent from the sender?
"What I send is what you see" principle :wink:

wp9015362:
As already mentioned in:

https://github.com/DrKLO/Telegram/issues/829

and:

https://github.com/DrKLO/Telegram/issues/222

please do not merge this.

Multimoon:
@wp9015362 this is a helpful change, it should be merged, and this clearly isn't a problem otherwise the powers that be would have standardized Emojiis for you know, everything else? Such as SMS/MMS.

desick:
I would love for this to be merged into the app. 
Please Please merge this, or at least add the option to use iOS versus Android.

wp9015362:
> @Multimoon wrote:
>
> this is a helpful change, it should be merged, and this clearly isn't a problem

It is not helpful and it clearly is a problem as he even wanted to force the native Emojis on users, see:

>  @robin0van0der0v wrote:
> 
> @midi Indeed, I think this is a much better solution to his problem than this one.
> That way we don't add a extra setting for solving such a small thing.
> 
> @DrKLO What do you think? A extra setting to use the native emoji set _or_
> always using the native emoji set when it is available?

So, as you can see, he even was thinking about forcing it instead of making it optional.

So much about "choice"...

nastys:
Why not merge? It's optional, after all. If you don't want to use system emojis, then just don't enable that option.
(it's much better than forking the app just to use system emojis)

wp9015362:
> @nastys wrote:
> 
> Why not merge? It's optional, after all. If you don't want
> to use system emojis, then just don't enable that option.

Even if I would not enable such an option, then still my chat partner could have enabled it and then his/her Emojis probably would look different than the ones I would be using. And when each chat partner has different Emojis it can lead to confusion.

It's all explained in the following article:

http://www.huffingtonpost.com/2014/06/27/emoji-meaning_n_5530638.html

nastys:
@wp9015362 I know some emojis may be different and that's why I think this should be disabled by default and possibly show a warning when enabling it. Using stock emojis has its advantages: for instance, they render faster (useful for old devices) and they look nicer with material design than Apple emojis. Anyway, Android Lollipop fixed some inconsistencies between the two sets and some devices use their own set.

wp9015362:
@nastys 

This is all being discussed over there:

https://github.com/DrKLO/Telegram/issues/829

Pox1994:
iOS emojis are so out of place in Material Design, they ruin user experience.

Walkman100:
https://play.google.com/store/apps/details?id=me.swineson.moegram.messenger

indrora:
Since this will get to people: @DrKLO has killed the issue tracker, hiding (nee destroying) history.

9peppe:
What. The. Hell.

I don't get why there are such strong feelings about this. There are at least 4 pull requests doing this in differing ways, one even includes an emoji pack selector instead of a switch to choose natives, and nonetheless, DrKLO does not want to merge this, replying "fork or die" because "apple did it first;" bullshit: ntt docomo did it first, and Unicode controls it.

How many telegram forks do we need? The official one has apple emoji, Telegram-FOSS does not have GMS, that I wanted. Do I need to make a THIRD one?

Multimoon:
Yes @9peppe , the general attitude from @DrKLO has been very defensive with the 'write your own and gtfo' attitude. As a developer with a few apps on the play store myself, if I user asks me to implement a option to allow for one of their preferences I always do, and I have yet to have any complaints of that. Even where Emojiis are involved.

thepaperpilot:
I would love to see the native emoji set used. Being native means it will automatically update when the system updates, is easier on everyone's device, and will make the app fit in more with the system its on. After doing so much to make telegram look like it belongs on each system, having universal apple emojis is a huge contrast. It also means people that change the system emojis because they prefer them will see their change reflected in telegram, which would be great. I don't think its a large issue if you can't see what the emoji will look like on the other users' end(s), because you are sending a unicode symbol, which is supposed to be separate from the implementation.

As others have said, if you are going to force a universal standard, it should definitely be one with an open license, like Noto.

Walkman100:
@9peppe you mean a fourth one, third one is: https://play.google.com/store/apps/details?id=me.swineson.moegram.messenger

tremby:
Since issues have disappeared I'll reiterate my comment on #829 here:

It seems the only real argument for using a single set of emoji on all platforms is so that everyone sees the same thing. But this is a false argument: this is an open protocol, and the person you're speaking to may not be using an official client. They might be in a text-only terminal, or using some web interface or bridge or any number of other possibilities. They might be using the official client but a newer or older version which displays the emojis differently.

So since there's no guarantee everyone sees the same thing I can't see any real argument against using the stock emoji set on each platform other than "I don't like this set". If "I don't like this set" is a common complaint, offer the choice. It seems very straightforward.

brownsmart:
Here is a question, why is the android version the only one where i cant change the colour of the emoji?

The0x539:
Because neither Android itself nor the app, with its nonsensically baked-in solution, support skin tone emoji yet.

brownsmart:
@The0x539 ... The app yes does not support it, but Android does because I can use it on WhatsApp, so I fail to see why it cant be done on Telegram too.

The0x539:
That app does support it.

brownsmart:
Your not making any sense... you previously said "Because neither Android itself nor the app, with its nonsensically baked-in solution, support skin tone emoji yet."

Which means neither Android or Telegram "the app" support it, now you saying the app does? ... so therefore what's the hold up?

The0x539:
It's supported by WA's implementation, but not that of Android or Telegram.

brownsmart:
Which translates to... not any time soon then? :joy:

brownsmart:
I stand corrected, Well done @DrKLO very greatful for the skin tone emoji addition. 👊

Multimoon:
I saw the new emojiis change and I got excited hoping that there were finally new emojiis or the ability to use your own. Then I saw it was just the useless skintone.

brownsmart:
@Multimoon useless to you, but to me a long awaited update. What other new emjoi's where you wanting?

Multimoon:
Hmm, maybe the ability to use stock system emojiis, like we've all been
begging and fighting for, for upwards of a year?

brownsmart:
If you refering to the stock android emoji, I always thought those were, boring and not fit for purpose...

The0x539:
The issue, though, is that that's an opinion.

brownsmart:
Good point.

private-static-void:
Having the toggle there for people who do want stock Android wouldn't affect anyone negatively. It should be added for the sake of those who have a preference for the Android ones.

tremby:
@robin0van0der0v, can you comment on the reason you closed this?

robin0van0der0v:
@tremby Because @DrKLO (or somebody else of Telegram) already made the decision more than a year ago to not implement this and hasn't responded since and because this pull request was pretty outdated anyway.

tjhorner:
There's also an alternative called Plus Messenger that allows this feature.

tremby:
Yeah, I'm aware there are a few. I'm using [Moegram](https://play.google.com/store/apps/details?id=me.swineson.moegram.messenger&hl=en) currently. But I'd still much rather this feature made it upstream.

PR #629, opened 2014-10-22
Change Search rule

ikicha:
past : startwith
now : contains


JunnanZ:
I think it's related to this issue about searching in Asian languages https://github.com/telegramdesktop/tdesktop/issues/564. 

@DrKLO Could you think of a way to reconcile that?

mrr010:
> It's wrong, people never search in name "Nickolai" substring "kol", they always search name from start.

Chinese language user always/only search like that because every Chinese character is a word by itself without spaces seperated. The current search functions is useless for us.

kmvan:
Need this feature.

spphinslove:
> 
> 
> It's wrong, people never search in name "Nickolai" substring "kol", they always search name from start.

Asia Languages Matter

JunnanZ:
People in the other thread keep saying this is a server issue. How is that related to this PR?

wizdzz:
Have you guys just think about add an option for us to switch this function?

PR #795, opened 2015-02-02
Fixed traslation for "Shared Media"

fduraibi:
The word "عدد" means number. Which doesn't match the english translation and also doesn't fit when displaying as action bar title.


DrKLO:
Thanks, new string added for this in last commit.

frunc:
I was waiting for this fix but it's not implemented right. I see new string when I enter drop down menu but when I choose it up there still stays the old shared media string

DrKLO:
@frunc yep, i forgot to change this string in one more place. Thanks, will be fixed in next version.

PR #800, opened 2015-02-02
Text Avatar for groups and users with no photo

fduraibi:
Two changes are proposed here:
1- Group names doesn't have a last name field, so if a group has a name of two words such as "Test Group" then Telegram for Android will only show the first letter T only (while Telegram Web and desktop show TG), so if there is no last name simply check if the first name has a space separating two or more words and use the first letter from the last word.

2- In languages like Arabic, letters that are have no space in between will usually link to each other, but in case of initials they should be separate, so I propose the use of "Zero-width non-joiner" character between the two letters which will make the letters display as stand-alone (non-joined) also it has no effect on other languages like English.

Image after applying the group name fix (groups now have two letters)
tel_joiner

Image after applying the Zero-width non-joiner, you see the effect is only on the Arabic letters but the English letters remain the same
tel_non-joiner

More info about Zero-width non-joiner can be found here:
http://en.wikipedia.org/wiki/Zero-width_non-joiner

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/Components/AvatarDrawable.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/Components/AvatarDrawable.java
@@ -176,7 +176,19 @@ public class AvatarDrawable extends Drawable {
                 }
                 lastch = lastName.substring(a, a + 1);
             }
-            text += lastch;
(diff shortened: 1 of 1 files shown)


DrKLO:
Thanks, added in last commit. Except for Android 2.x, it's displaying strange character for this "Zero-width non-joiner" character.

PR #810, opened 2015-02-05
Fix for GIF search when using Arabic interface

fduraibi:
Without "Locale.US" the String.format will change the numbers 0 & 53 to the device local language representation. For example, if the Telegram language is set to Arabic the formatted URL will have the offset and count in Hindu-Arabic numerals representation like the picture below

tel_gif_before

The fix will insure the numbers stays as is

* Without the fix the search result is always empty!


DrKLO:
Thanks, added in last commit.

PR #812, opened 2015-02-07
Update RemoveAllFragments()

PrintHelloWorld:
Remove the a-- makes no sense.

public void removeAllFragments() {
        for (int a = 0; a < fragmentsStack.size(); a++) {
            removeFragmentFromStack(fragmentsStack.get(a));
            a--;
        }


PrintHelloWorld:
My mistake, I didn't think about this rearranging the stack on each removal.

PR #816, opened 2015-02-08
.nomedia file in cache

4m4rOk:
On some phones (e.g. HTC M8) gallery apps (e.g. HTC Gallery) show the Telegram cache folder as an album. 
Even though this is not the fault of the telegram app I believe for the UX it would be better if the app created a .nomedia file, so the (not so perfect) gallery apps skip scanning the cache folder.

Thanks to @asdofindia for the code.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/FileLoader.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/FileLoader.java
@@ -81,6 +81,9 @@ public class FileLoader {
         try {
             if (!dir.isDirectory()) {
                 dir.mkdirs();
+            if (type == MEDIA_DIR_CACHE){
+            new File(dir, ".nomedia").createNewFile();
+            }
             }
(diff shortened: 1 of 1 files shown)


Lapayo:
+1 :thumbsup:

DrKLO:
You should turn off "Save to gallery" in app settings. It will automatically create .nomedia file.
https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/java/org/telegram/android/MediaController.java#L1795

4m4rOk:
@DrKLO Thanks for your quick reply. This only creates .nomedia files in Telegram Audio, Telegram Documents, Telegram Images and Telegram Video, correct?

I encountered the issue of some phones showing /Android/data/org.telegram.messenger/cache in their gallery apps. It would be great if you could also add a .nomedia file in there. ))

4m4rOk:
@DrKLO Thanks for the quick fix! )

PR #828, opened 2015-02-14
Add plural rules for Aragonese and Asturian languages

danielmartinez:
Telegram doesn't find the plural forms for Aragonese and Asturian, so it shows "LOC:ERR" messages when plural strings are used, so defining this rules for both languages makes it to show the correct strings.


DrKLO:
Thanks, added in last commit.

PR #933, opened 2015-03-23
Better tg:// uri support

LucaFraga:
Cleaned up the code and added msg (or send / message) for sending a
message via url.


LucaFraga:
The code was tested and works fine, if this change gets approved i will add the same code on the other clients asap.

averinasvetlanapavlovna-lab:
OOO r class code

Issue #1005, opened 2015-04-14
Share something from another app


benkio:
Hi,
I found a problem in the app when i try to share some file from another app, telegram show a toast with the "attatchment unsupported" message.
In particular i notice, debugging the app, than this happens because in the prepareSendingDocumentInternal method the path and the originalPath are null and so in the copyDocumentToCache method an FileNotFoundException is throw.

I try it with the app of google drive for instance. If you make a file without download it on the smartphone and share it directly instead, the problem descripted above happen.
Unfortunately i had some problem in the setting up my google accout with the emulator. I made the same test with an app that share some sounds. See attatchments images
I test it on master, righ now
![schermata 2015-04-14 alle 09 24 19](https://cloud.githubusercontent.com/assets/4698396/7132354/b60de4ce-e288-11e4-964f-bceb7f505579.png)
![schermata 2015-04-14 alle 09 24 43](https://cloud.githubusercontent.com/assets/4698396/7132356/b6311a3e-e288-11e4-8d07-00f70837f79b.png)
![schermata 2015-04-14 alle 09 25 30](https://cloud.githubusercontent.com/assets/4698396/7132355/b61a4bec-e288-11e4-80d9-a813ab069225.png)

Sorry for my english mistakes :)
Best regards

ps: with whatsapp it works.
pps: it will be awesome if telegram also recognize that the file shared in this way is a media file and share it like a video/audio. Like when an audio message is recorded and sended.

AUDI-A4:
Same problem found while sharing videos link from external app like youtube.It does not show the link preview of that video which is shared by external app Youtube.In whatsapp, it does nicely and show preview too.

tsubus:
#1008

PR #1008, opened 2015-04-14
Always open chat before sending text message shared from other app

tsubus:
* This allows to edit the reply text and/or show the link preview

Change-Id: I2f445e554d72da029e49d65623aa7b87a543ab05


tsubus:
#1005

PR #1013, opened 2015-04-15
Add default text answers

Vahn84:
for Android Wear Notifications


robin0van0der0v:
Looks good. Maybe some more replies?

Vahn84:
I was out of ideas

would you suggest me some useful answers?

Bikonja:
The most sane suggestion would be an entry in the settings where every user could enter his/her own replies for Android wear. I personally love when there are just basic ones out of the box and I can really personalize my responses, otherwise there's no point in scrolling through a bunch of replies I never use only to get to the replies I added myself. Of course, this means more code... Otherwise, I absolutely support this and even like this, I would be happy if the pull request is merged into Telegram.

Vahn84:
@Bikonja 

considered your request, and i'm adding that feature. I will open a new pull request

PR #1027, opened 2015-04-22
Fix Emoji Recents Behavior

dbt0210:
Currently, selected emojis are always added to the beginning of the recents list. This shifts the whole layout, making it hard for the user to memorize it.
Selected emojis should be added at the end of that list instead. They are only moved to the beginning if the user selects them again via the regular emoji grid. Also, we make sure to add them in the second last slot, so they replace the last emoji if we already have the max number of emojis in there.
This makes the recents layout more consistent and easier to remember for the user.


ansgarm:
👍

Issue #1039, opened 2015-04-30
Option to not edit photos


Geobert:
So we can save a tap before sending a pic

Mindstormer619:
... Seriously?

Geobert:
Yes, I'm serious, just added the needed code

Mindstormer619:
> save a tap

¯\_(ツ)_/¯

PR #1121, opened 2015-05-26
option to bypass photo viewer on taken photo

Geobert:
fix for  #1039


ADTC:
Could you also make the photo viewer appear when an image file is shared from another app? For example a screenshot. The photo viewer only appears when I take a picture with the camera, allowing me to crop, edit and add a title. I also want to be able to do the same when I share an image from somewhere else (not the camera) - to crop, edit or add a title.

PR #1141, opened 2015-06-04
Telegram on android M crashes everytime the Emoji popup is shown

ClemensElflein:
As the PopupWindow.getContentView().getLayoutParams() in Android M doesn't return a WindowManager.LayoutParams instance anymore, but a FrameLayout.LayoutParams, the cast would induce a foce close.

Therefore I rewrote the code in order to avoid casting the params (therefore being compatible with older versions of android).

It would be great if you'd accept my pull request. Then I could use the official Telegram App again. I'm sure that multiple users have this problem.


tsubus:
When @DrKLO uploads the 3.2.4 update sources it should all be fixed. It's been a while now since he posted it on the Play Store, so I imagine it won't take that long.

dkaraush:
Yep, I see it was merged, closing

PR #1198, opened 2015-06-27
Readability

pjrobertson:
I'm hoping to make some additions to the drawer in the future, but at first I thought it was worth improving the readability of the file.

I have changed to using constants for the view types and also to using a switch - using a switch should make it easier to add new rows in the future (and it is also faster)


yazeed44:
You da real mvp !!
They should really clean their code a little

danialbehzadi:
It's great

7ShaYaN7:
Hope to merge ASAP.
Good idea, I really thought on the switch statements instead of if/else glued together.

I wanna do this optimization for whole of the project classes if @DrKLO mind this problem and make it a merge at last?!

pjrobertson:
See #1386 for an updated PR with merge conflicts resolved

PR #1203, opened 2015-06-29
Editable Android Wear Response

Vahn84:
Implemented a new activity inside the notifications settings to manage the android wear default text answers


Royal2000H:
Would love to see Telegram increase Android Wear support. For example, having the ability to open the Telegram chat list from the watch, rather than just being able to reply. More reply options is a great start though.

huxaiphaer:
Please resolve these conflicts.

toivan:
Android Wear OS version required

PR #1231, opened 2015-07-18
Added input text size option to settings

AlgimantasMilasius:
This messenger is one of the best i've used. The only thing I don't like about it is that i am not able to change the size of the input text. The default size is too big for me (I don't know why but it just annoys me) so I decided to add an option in the settings menu to change input text size the same way as you would change output messages text size.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/android/MessagesController.java
+++ b/TMessagesProj/src/main/java/org/telegram/android/MessagesController.java
@@ -99,6 +99,7 @@ public class MessagesController implements NotificationCenter.NotificationCenter
 
     public boolean enableJoined = true;
     public int fontSize = AndroidUtilities.dp(16);
+    public int inputFontSize = AndroidUtilities.dp(18);
     public int maxGroupCount = 200;
     public int maxBroadcastCount = 100;
     public int groupBigSize;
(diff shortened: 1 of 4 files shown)


IgnasD:
:+1:

TMessagesProj/src/main/java/org/telegram/android/MessagesController.java
```diff
@@ -99,6 +99,7 @@
 
     public boolean enableJoined = true;
     public int fontSize = AndroidUtilities.dp(16);
+    public int inputFontSize = AndroidUtilities.dp(18);
```
blindMoth: avoid copy paste magic numbers - declare const for default font size int

TMessagesProj/src/main/java/org/telegram/ui/SettingsActivity.java
```diff
@@ -384,6 +386,31 @@ public void onClick(DialogInterface dialog, int which) {
                         }
                     });
                     showDialog(builder.create());
+                } else if (i == inputTextSizeRow) {
+                    if (getParentActivity() == null) {
+                        return;
+                    }
+                    AlertDialog.Builder builder = new AlertDialog.Builder(getParentActivity());
```
blindMoth: call getParentActivity() once and store in a local var

TMessagesProj/src/main/java/org/telegram/ui/SettingsActivity.java
```diff
@@ -1042,6 +1069,10 @@ public View getView(int i, View view, ViewGroup viewGroup) {
                     SharedPreferences preferences = ApplicationLoader.applicationContext.getSharedPreferences("mainconfig", Activity.MODE_PRIVATE);
                     int size = preferences.getInt("fons_size", AndroidUtilities.isTablet() ? 18 : 16);
                     textCell.setTextAndValue(LocaleController.getString("TextSize", R.string.TextSize), String.format("%d", size), true);
+                } else if (i == inputTextSizeRow) {
+                    SharedPreferences preferences = ApplicationLoader.applicationContext.getSharedPreferences("mainconfig", Activity.MODE_PRIVATE);
+                    int size = preferences.getInt("input_fons_size", 18);
```
blindMoth: consider customizing per screen size

martinszy:
Hi, I'm talking to a new user that refuses to use the app because it does not respect the systemwide text setting, is anyone interested in following up with this PR? 
It seems like the comments are not very hard to fix.

PR #1244, opened 2015-07-24
Update PasscodeView.java

dkim91:
Incomparable type. Need to type cast


DevendraSingh91:
error in passcodeview.java
Error:(996, 46) error: incomparable types: Object and int

PR #1284, opened 2015-08-10
Add Chinese value

iwanderer:


auchri:
Translations are moved to transifex: https://www.transifex.com/telegram/telegram/

iwanderer:
Sorry...Get it !

PR #1303, opened 2015-08-25
mute a chat with custom date time

fansanelli:
As written in subject, using Android pickers for date and time, users can choose a custom date time until when the mute is valid


Playez:
why bro

Playez:
such a good conmit

fansanelli:
@Playez it have been opened since 2015..

PR #1351, opened 2015-09-16
Add a feature: When the keyboard or "popup" is showing, then click on the chat list, people can hide the input keyboard or "popup"

drakeet:
Add a feature: When the keyboard or "popup" is showing, then click on the chat list, people can hide the input keyboard or "popup";

Preview GIF: 
demo

It is easier to use, more intuitive.

I sincerely hope that it can be adopted, thank you very much!


drakeet:
*It's useful to close emoji views when in ChatActivity.*

default

PSeitz:
+1, would like that too

similato:
+1

PR #1363, opened 2015-10-09
BUG in notification settings activity

7ShaYaN7:
[BUG] after call this method, when MessagesLed and GroupLed row fit in a same screen both of them display in a same color.

Because of Issues section is disabled I had to propose that here.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/NotificationsSettingsActivity.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/NotificationsSettingsActivity.java
@@ -390,7 +390,8 @@ public class NotificationsSettingsActivity extends BaseFragment implements Notif
                                 editor.putInt("GroupLed", colorPickerView.getColor());
                             }
                             editor.commit();
-                            listView.invalidateViews();
+                            listView.invalidateViews();  /* [BUG] after call this method, when MessagesLed and GroupLed row
+                                                            fit in a same screen both of them display in a same color */
                         }
                     });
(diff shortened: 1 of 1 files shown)


7ShaYaN7:
Trying with listadapter.notifyDataSetChanged() get the same results, too

7ShaYaN7:
Please turn back Issues section

7ShaYaN7:
For more clearance, in a TextColorCell view type group, the last TextColorCell view force other TextColorCell view items to make their color sample as the last one.

After clicking on the other TextColorCell items you notify that their color samples didn't change and only the last one determine color sample of all TextColorCell views.

7ShaYaN7:
Please be helpful to solve this bug.

I would be pleasant to present more details about this issue, if you want

7ShaYaN7:
photo_2015-10-09_14-46-28

srazi:
Why issue tracker is disabled? Where we should report bugs?

7ShaYaN7:
Issue tracker is a very useful part of each open source project. @DrKLO devs like me wants to be more connected with the project that they're developing upon it.

joshery420:
I agree with @srazi and @7ShaYaN7. It's insane that the Issue Tracker is disabled for an "open source" project.
We need a way to make issues known other than by fixing them ourselves and submitting pull requests. Not everyone is a developer here.

7ShaYaN7:
the problem solved by the Dev. thanks in advance but wish to see issues again!

PR #1365, opened 2015-11-05
Increase number of rendered emoji to 200

flode:
I need to display more emoji for a bot game.


DrKLO:
No, it will cause performance problems.

nakami:
seven years have passed and modern devices should be capable of an increased limit. might you reconsider this?

image

PR #1366, opened 2015-11-16
Fix markdown syntax

jonan:
This headline has wrong markdown format. Github displays it correctly, but some other renders wont.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [Telegram](http://telegram.org) is a messaging app with a focus on speed and security. It’s superfast, simple and free.
 This repo contains the official source code for [Telegram App for Android](https://play.google.com/store/apps/details?id=org.telegram.messenger).
 
-##Creating your Telegram Application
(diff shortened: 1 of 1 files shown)


tjhorner:
As in? The `##header` syntax should conform to the [Daring Fireball Markdown Spec](http://daringfireball.net/projects/markdown/).

jomo:
Actually the Daring Fireball Spec [says](http://daringfireball.net/projects/markdown/syntax#header) that these are Atx-style headers.

CommonMark [notes](http://spec.commonmark.org/0.21/#atx-header):
> At least one space is required between the `#` characters and the header’s contents, unless the header is empty. Note that many implementations currently do not require the space. However, the space was required by the [original ATX implementation](http://www.aaronsw.com/2002/atx/atx.py) […]

So if the spec says these are atx headers, and atx requires spaces, then the spec indirectly requires spaces, I guess :sunglasses:.

PR #1367, opened 2015-11-23
Simplify `searchForHttpInText` method

mrcasals:
This PR simplifies code logic and makes it more comprehensible.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/ChatActivity.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/ChatActivity.java
@@ -2135,30 +2135,8 @@ public class ChatActivity extends BaseFragment implements NotificationCenter.Not
     }
 
     private boolean searchForHttpInText(CharSequence string) {
-        int len = string.length();
-        int seqLen = 0;
-        for (int a = 0; a < len; a++) {
-            char ch = string.charAt(a);
-            if (seqLen == 0 && (ch == 'h' || ch == 'H')) {
-                seqLen++;
-            } else if ((seqLen == 1 || seqLen == 2) && (ch == 't' || ch == 'T')) {
-                seqLen++;
-            } else if (seqLen == 3 && (ch == 'p' || ch == 'P')) {
-                seqLen++;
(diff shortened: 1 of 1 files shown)


txus:
Actually I'd say this PR implements the code for the first time, without needing a hand-written parser with manual backtracking and off the charts cyclomatic complexity. Kudos!

jomo:
What the hell… :+1:

TMessagesProj/src/main/java/org/telegram/ui/ChatActivity.java
```diff
@@ -2135,30 +2135,8 @@ public void onClick(DialogInterface dialogInterface, int i) {
     }
 
     private boolean searchForHttpInText(CharSequence string) {
```
jomo: You could also rename `string` to `sequence` or some other more accurate name, while you're on it.
`string.toString()` looks kinda weird…

tjhorner:
Holy shit, that code was actually in prod? That's probably the worst I've ever seen.

Good job @mrcasals

3mph4515:
Maybe someone want to make some perfomance tests? :1234:

DrKLO:
It was done like this to avoid memory allocation on every entered character, so what's the problem?
Anyway it's removed and replaced with regex and moved to background thread.

jomo:
@DrKLO [that regex](https://github.com/DrKLO/Telegram/blob/74def22213846b1f90a26c86670035079dea032c/TMessagesProj/src/main/java/org/telegram/messenger/AndroidUtilities.java#L97-L123) doesn't match all URLs, for example [`http://to`](http://to). I don't expect "to" to be parsed as a link, but any valid URL prefixed with `http://` should match, no?

jomo:
Additionally, it only seems to match `EXAMPLE.COM` for the string `HTTPS://EXAMPLE.COM`.
I guess that would end up being rendered as _HTTPS://[EXAMPLE.COM](http://example.com)_, effectively dropping the https protocol?

DrKLO:
@jomo I'm using regex from android Linkify. I don't think that somebody will write http in upper case.
This regex used only for webpage preview request from server. For links highlighting we using server side text parser (cloud chats only).

PR #1368, opened 2015-11-27
Include missing unistd.h header.

Bubu:
On Linux unistd.h is needed for fsync().
The native library does not compile otherwise.


DrKLO:
Thanks

PR #1370, opened 2015-12-01
Started localization of country list

Minty123:
Moved countries.txt from assets to raw folder to make it translatable and translated it into German. I might be the only person bugged by that but it always annoys me when a system is only partially translated... :smile: Greetings!


robin0van0der0v:
Maybe it would be better if all countries would be shown in only the language of those countries? This would save lots of translation work.

Minty123:
@robin0van0der0v When I did it, I thought the language selection of Android would be in the system locale as well. Seems like I was wrong though.... :sweat_smile: You're right of course! Guess I'll just close this then and maybe start looking for the country names in their respective language.

PR #1371, opened 2015-12-01
Comparisons of incomparable types

SHi-ON:
All the matter is clear in the files!


3mph4515:
Try to use Java 7 not 8

7ShaYaN7:
But this issue can be solved by a tiny change. so why not do it?!

3mph4515:
There is no issue. Everything is ok
It's your problem that you are using J8

PR #1372, opened 2015-12-11
Fix non-bmp first character split issue

ebraminio:
If you set your last name to a name started with an emoji (as a non-bmp character) you will get invalid character shape. This _UNTESTED_ patch tries to resolve it by incorporating offsetByCodePoints for splitting first character instead. Someone should check if this fix the issue before the merge of course, I just speculatively patched it with a tablet as my lack of access to a suitable for development machine so pardon me before hand if it has any issue, I guess this at least leads to an actual fix in case itself couldn't.

Steps to repro the issue this patch tries to fix:
1. Set your last name or first to an Emoji (or any non-bmp characters, say a historical character)
2. Open the drawer of the app and look at your text avatar, if you didn't set an avatar yet of course
Actual:
You see something like an unrecognized character as first character splittion failed.
Expected (with the patch):
You don't see anything weird

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/Components/AvatarDrawable.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/Components/AvatarDrawable.java
@@ -167,6 +167,12 @@ public class AvatarDrawable extends Drawable {
         color = value;
     }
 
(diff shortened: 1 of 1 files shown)


ebraminio:
Didn't test it but it seems is resolved anyway

PR #1373, opened 2015-12-14
Edit korean SentSmsCode string:unnecessary br tag

drzix:
Although the strings of other languages in SentSmsCode don't
include '<!CDATA..', the Korean string does that.
Unfortunately, that string is not escaped by replaceTags().
So, the CDATA tag must be deleted.

korean_br_err


zhufui:
看起来像是韩文

DrKLO:
Thanks.

PR #1384, opened 2016-01-04
Make phone numbers in media captions clickable

astarcev:
It's very annoying that I can't tap a phone number in a picture caption (which is possible in regular messages).
This should make these phones clickable.


DrKLO:
Thanks.

PR #1385, opened 2016-01-04
Increase limit to linkify phone numbers from 100 to 200

Kondra007:
100 symbols is too few to make sending phone cards (not telegram users' cards, but any phone cards).


vbolgov:
This is great idea! waiting for it!
for creating cards with different info and using it (sharing card for example)

DrKLO:
Thanks.

albo94:
Please increase the limit to 400 or 500....it could be very helpful!!!
Thank you very much!!!

PR #1391, opened 2016-01-31
To recognize wav and aac files being sent from other apps as audio instead of file attachment

iori57:
When other apps sends a wav or aac audio clips using intent, because its duration cannot be determined it is being attached as file attachment. With this change they can be played using internal media player without issue.


iori57:
Another thing I notice is that if you select attachment and then Audio, all the audio files (including .wav) in my file system are able to display properly with the right duration, and selecting them will instantly send them as media files (can be played with internal player). If I use other file manager and share the same .wav audio file to telegram, it is recognized as file attachment... so clearly there is an inconsistent behaviour between attaching audio file within chat with attaching audio file via 'share to' from other apps

PR #1392, opened 2016-02-01
Location sharing enhancements

Bubu:
This PR adds two independent, small enhancements:
-  Add sending location objects via geo: message. 
  This patch add the ability for Telegram to parse locations form messages containing a `geo:<lat>,<lon>,<zoom>` string.
  This is used by various map applications including the popular OSMAnd and Orux maps. This patch enables the user to share the position of a POI or destination with Telegram and have it send as a location.
-  Add support for opening locations without gmaps.
  When google maps is not found on the device, try to open a location message with a `geo:` intent so it can be interpreted by any other maps application.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/SendMessagesHelper.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/SendMessagesHelper.java
@@ -3614,6 +3614,29 @@ public class SendMessagesHelper implements NotificationCenter.NotificationCenter
         });
     }
 
+    public static void prepareSendingLocation(final Location location, final long dialog_id) {
+        MessagesStorage.getInstance().getStorageQueue().postRunnable(new Runnable() {
+            @Override
+            public void run() {
+                Utilities.stageQueue.postRunnable(new Runnable() {
+                    @Override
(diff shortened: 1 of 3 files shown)


Bubu:
Rebased on top of latest available source.

TMessagesProj/src/main/java/org/telegram/messenger/SendMessagesHelper.java
```diff
@@ -3614,6 +3614,29 @@ public void run() {
...
+                                SendMessagesHelper.getInstance().sendMessage(mediaGeo, dialog_id, null, null, null);
+                            }
+                        });
+                    }
+                });
+            }
+        });
+    }
+
     public static void prepareSendingPhotos(ArrayList<String> paths, ArrayList<Uri> uris, final long dialog_id, final MessageObject reply_to_msg, final ArrayList<String> captions, final ArrayList<ArrayList<TLRPC.InputDocument>> masks, final InputContentInfoCompat inputContent, final boolean forceDocument, final ArrayList<Integer> ttls) {
```
Jhuff6: #1461

PR #1393, opened 2016-02-01
Show menu item "Share my contact" for users in contact list.

fiftin:
For unknown reasons, the menu item "Share my contact" is shown only for users not in the contact list. If the user is in my contact list, that does not mean that I am on his contacts list


Gollam:
Why hasn't this been merged in 3 years? This is very annoying, I just ran into this.

PR #1398, opened 2016-02-28
Panic Code: clear all secret chats with specific passcode

tfoldi:
With this patch one can specify a special `panic code` that clears all secret chats history (both locally and remotely). This is pretty useful if you're forced to unlock your application since you can remove all proofs before things are getting hot. 

**How does it work?**

Panic Code can be set in _Passcode Lock_ menu under settings. If Panic Code is enabled and you use that code instead of pass code then all secret chats are cleared. Panic Code does not unlock the application, you still have to use your existing passcode.

**Technical stuff**

I added a new dialog type to `PasscodeActivity` with `type==3`. That will reuse the existing passcode UI elements. 

Please let me know if I need to change anything to get this PR approved. Thx.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MessagesController.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MessagesController.java
@@ -548,6 +548,10 @@ public class MessagesController implements NotificationCenter.NotificationCenter
         return chats.get(id);
     }
 
+    public ConcurrentHashMap<Integer, TLRPC.EncryptedChat> getEncryptedChats() {
+        return encryptedChats;
+    }
+
     public TLRPC.EncryptedChat getEncryptedChat(Integer id) {
         return encryptedChats.get(id);
     }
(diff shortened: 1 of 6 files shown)


tfoldi:
I just updated the PR to use version 3.6.1

tfoldi:
@DrKLO , any chance for merge or feature import? if not, I'll close it 😞

richin13:
Nice feature

z3ntu:
Great idea!

chtisgit:
I like this idea, but in my eyes the panic code should also unlock the app.
So it really looks like you unlocked it at first try.

tfoldi:
@chtisgit the idea was that if the app is locked, then you type this code that deletes everything but keeps the app locked. then you decide if you'd like to show/type the right pin and unlock it or keep it locked while all local history is removed.

however, I can make this panic code an unlocker code as well. I just don't see that how this feature will be merged, it was opened a year ago and there is still no response from @DrKLO

chtisgit:
@tfoldi I actually didn't look at the year and thought you did this a few days ago...
It's really an unfortunate situation with the maintainance of this app.

Giando:
Beautiful idea, and great feature!
Why was it made to fall into forgetfulness?

Ravshanchi:
Kod kelishi sekin va natoģri

PR #1406, opened 2016-04-14
Fix clearItemsMethod at ActionBarMenu 

abcdeiko:
Old `clearItems` method at `ActionBarMenu` doesn't clear view correctly.
Each iteration remove one `View` from parent `View`, that is a `getChildCount` method take incorrect value.
Some example:
getChildCount = 4;

Some example:
a = 0, a<**4**, removeView (getChildCount = getChildCount-1)
a = 1, a<**3**, removeView
a = 2, a<**2**,, removeView
exit cycle.

*_The last menu item isn't removed_*


DrKLO:
Thanks!

PR #1412, opened 2016-06-03
Reducing the function calls by simply creating a variable

aravindkanna:
Here in the `ActionBarMenu.java` the function getChildCount() is being called in the `for` loop in all iterations. Instead creating a variable can make this simple I guess.


build.gradle
```diff
@@ -6,6 +6,6 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.1.0'
-        classpath 'com.google.gms:google-services:2.1.0'
+        classpath 'com.google.gms:google-services:2.1.0-alpha3'
```
abcdeiko: This change doesn't match with your commit message.

abcdeiko:
@aravindkanna good job!
But you don't forget one of base rule: "One branch -> one feature". Therefore, you don't need to change gradle dependencies at this commit, i think.

aravindkanna:
so what should I do now? Create a new branch or change the current branch? Actually that was happened when I tried to merge. The fault was to commit before pulling the current remote.

abcdeiko:
@aravindkanna you don't need any action now, because **DrKLO** merged you commit.
To prevent merge conflicts you can fetch last changes from master branch and rebase it on your working branch before create a new pull request.

similato:
Thank you.

PR #1424, opened 2016-10-25
Fixed a memory leak

mwajeeh:
Usually `onVideoTakeCallback` is anonymous class in Fragment which causes memory leak when Fragment is no longer active. Setting onVideoTakeCallback to null when its no longer required fixes this issue.


dkaraush:
Thanks! As I see in code, it is already is setting to null, I believe it was merged

PR #1435, opened 2017-04-04
Fixed the second header appearance in README

rosenpin:
Fixed the "Creating your Telegram Application" header.
Some browsers would show ## before the "Creating your Telegram Application".
This should now be fixed


theel0ja:
Same as https://github.com/DrKLO/Telegram/pull/1433

PR #1436, opened 2017-04-07
Fix subheader in README

rogersachan:
GitHub has moved to cmark and broke some old GFM style in markdown. This PR fixes that.


theel0ja:
Same as my #1433

PR #1447, opened 2017-07-14
Add a setting for using system emoji

39bit:
Some leftover commented code uncommented, as well as new code added to make the emoji panel work properly.

A translation string already exists, so adding a setting was trivial.

The emoji panel uses scaling with an oversized font, the size of which is measured and then scaled accordingly.

Under replaceEmoji, the MessagesController.getInstance().useSystemEmoji if condition was moved lower down to allow links and other formatting to still work even if the setting is enabled.

The allowBigEmoji / draw single big emoji setting remains unimplemented in this PR.


nyuszika7h:
Does this fix the bug with `useSystemEmoji` breaking entities (links, mentions, formatted text all show up as plain text) that exists since build 1026?

39bit:
@nyuszika7h Yes, see:

"Under replaceEmoji, the MessagesController.getInstance().useSystemEmoji if condition was moved lower down to allow links and other formatting to still work even if the setting is enabled."

NatiSD:
I found a bug at the new beta update.
When you open the photo editor and writing on the pic. If hou pressing on the back button to delete the app is freezing and not working.

And also i have seen sync problem in the bio on the url description of massages.

39bit:
@NatiSD What does that have to do with this PR?

PR #1458, opened 2017-12-18
CameraView: add suport for 16:10 aspect ratio

Saylance:
Fixes preview glithces on device with aspect ratio 16:10 (ex. 1280x800)

For example, my Galaxy Note (N7000), with HD display has resolution 1280x800 (the normal resolution for HD is 1280x720), and aspect ratio 16:10 (the normal is 16:9).

Telegram preview view, has glitch. Looks like colors marix is smaller then real photo. (see pic)
(Photo that i took with Telegram in-camera, is OK, no problem with them, glitches only in preview)

First pic (glitches in preview):
https://imgur.com/8WNO77c

Second pic (photo that it tooks):
https://imgur.com/OykIEoP

I made a patch to fix this bug, but i cant test it.
https://github.com/Saylance/Telegram/commit/5384210471a2e3a3f868c99fb65adda9078242c4

And generaly, its pretty crabby code, becouse now we have many aspect ratio of screens "Infinity" (trends of 2017), and we (have?) to manually hardcore them there?

(Anyway, maybe patch isnt work, i cant build this (without normal PC))
Thanks for help.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/camera/CameraView.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/camera/CameraView.java
@@ -128,7 +128,7 @@ public class CameraView extends FrameLayout implements TextureView.SurfaceTextur
             return;
         }
         float size4to3 = 4.0f / 3.0f;
-        float size16to9 = 16.0f / 9.0f;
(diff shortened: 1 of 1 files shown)


dkaraush:
Now camera view properly scales texture into surface, thanks for the PR!

PR #1467, opened 2018-03-01
For privacy reasons, we should give priority to internal storage cache

eugeniobaglieri:
Cache directory contains all thumbnails and image (also from secret chat). So we should check first if a private internal storage is available, and, if not, fallback to external.
We should save them in internal storage instead of publicly accessible external storage cache.


sphinxy:
It makes sense to use ExternalCache for normal chats, those data is not very important normally and since cache for telegram typically is big, external storage is preferable.

But for secret chat it really mandatory to use internal cache which is not accessible by any other apps. Otherwise, your secret image of boobs can be viewed by any apps.

sidahmed-malaoui:
This methodology is most. Telegram should use a private storage for cache.

PR #1471, opened 2018-04-27
Update CameraView.java

NooAn:
method initCamera() have bool parameter and don't use


dkaraush:
I see it is different code now, thanks for the PR

PR #1475, opened 2018-07-11
Update OpenChatReceiver.java

ALIRESA:


dkaraush:
do you want to crash? why why why

PR #1477, opened 2018-08-18
Correct avatar_nameInMessage for violet and pink

gagarski:
Previously these colors were set to blue (same as `avatar_nameInMessage_blue`) in all default themes which is frustrating.

The chosen colors were calculated as (round(0.95 * r), round(0.95 * g), round(0.95 * b)) for corresponding `avatar_background` and tested using Telegram built-in theme editor with all three default themes and seem to look fine.


gagarski:
It's now fine for all themes except for Default.

Jahondor:
Kim bu

ср, 5 июн. 2019 г., 17:08 Kirill Gagarski <notifications@github.com>:

gagarski:
> Kim bu

?

PR #1482, opened 2018-09-25
Bugfix: correctly render vcard of a selected inline query result

acran:
Currently when selecting a contact as inline query result which contains a vcard the vcard is not rendered at first.
After switching between chats the vcard is rendered correctly though.

**How to reproduce:**
Use a bot which returns contacts as inline results and select a contact - for testing you can use [@InlineVcardBug_bot](https://t.me/InlineVcardBug_bot), source [here](https://github.com/acran/InlineVcardBug_bot):
1-inline-query

**What happens:**
 the selected contact is sent into the chat but the contained vcard is not rendered:
2-broken-rendering

When leaving the chat and opening it again the vcard is displayed correctly:
3-correct-rendering

**What should happen:** the vcard gets rendered right away.

This pull request fixes this by correctly passing down any contained vcard.


mdan312:
I have just updated Telegram from Android and cannot anymore find the way to attach a contact? (send a VCARD as an attachment)

PR #1487, opened 2018-12-04
Create SSR-PROXY

stupicode:
Any idea for new proxy option something like SS or SSR , Iran block http ports and noise on the MTPROTO


dkaraush:
Please leave suggestions at https://bugs.telegram.org/?tag_ids=2&type=suggestions&sort=time, thanks!

PR #1498, opened 2019-03-28
Bugfix: Added correct match check of phone numbers.

23rd:
Currently it is impossible to login second account with another phone number which is the same phone number you already used but with extra digit at the end.
For example user can have the following accounts:
- +0 1234567890 (10 digits)
- +0 12345678901 (11 digits)

In another clients (e.g. Telegram Swift) user will be able to login both accounts in the same time.

The code of Pull Request for checking phone numbers uses the built-in [Android function](https://developer.android.com/reference/android/telephony/PhoneNumberUtils.html#compare(java.lang.String,%20java.lang.String)), which was added to the API level 1.


23rd:
Oh, I see.
Seems like it's better to fix this manually with your own hands instead of just taking a pull request.
Nice. =)

PR #1499, opened 2019-03-31
Improvements to README.md

MarcAnt01:


MarcAnt01:
Thanks for approving, but according to all the pr section, this repo is just a mirror of the source code, there's no real contribution of external people

BO41:
Why does the description not say "mirror"?
@DrKLO could you change the description to "Telegram for Android source. read-only mirror that isn't used for accepting contributions" or something along these lines, so that people can know? :smiley:

MarcAnt01:
> Why does the description not say "mirror"?
> @DrKLO could you change the description to "Telegram for Android source. read-only mirror that isn't used for accepting contributions" or something along these lines, so that people can know? 😃

In the past I have seen them to merge some pr, but just few. I think not having a single mention is not transparent with users.
@DrKLO any thoughts about that?

PR #1508, opened 2019-06-18
Fix wrong empty view on gif searching view

fisher3421:
Fix showing empty view for gifs searching result when there is one item

Before

After


dkaraush:
should be fixed, thanks for PR!

PR #1510, opened 2019-06-18
Fix linkify email in messages and caption

4xes:
Fix linkify email after sending a message or image with caption.

Before
gif_mail_before2
After
gif_mail_after


dkaraush:
Code was moved from Linkify to own solution, so this PR wouldn't work now. But issue is relevant, would look into it.

PR #1525, opened 2019-08-22
Theme accents

alexvasilkov:
* **Tinting**: I tried to come up with tinting formula that would better preserve contrast between texts and backgrounds. Though black accent for Dark theme or white accent for Arctic theme will still make texts unreadable.
* **Perfromance**: There is no performance impact if default (blue) accent color is selected. In case of other accent colors there will be imperceptible extra load during theme initialization (1ms on Samsung S10+, 9ms on old Nexus 5), but during regular app usage there will be no extra load.
* **Fixed**:
  * Accent color wasn't correctly tinted in few places. E.g. message input cursor, new group image background, etc.
  * Text handling controls used blue color set in Android theme, I changed them to neutral grey to not conflict with accent colors.
  * Opening Theme Preview screen was rewriting local user's theme with same name, even if Cancel button is pressed.
* **Migration**: Users who used old Dark theme (either as current or as night theme) will be switched to Dark Blue theme with grey accent after update, since it is quite close to original Dark theme. As alternative we can save old Dark theme as user's custom theme to keep it exactly the same.
* **Theme sharing**: When sharing Dark or Arctic theme with accent color applied all colors will be saved as already tinted.
* **Feature suggestion**: We can allow using different accent colors for different user accounts, so that the user can easily understand which account he is currently under just from accent color. We can save theme accent to account-specific perefences right away, or add extra switch like "Use separate colors for different accounts", though it may complicate Settings screen UI.
* **Known issue**: PhotoViewer does not use themes mechanizm, all colors are hardcoded there, so it won't be tinted with accent color. To fix this we'll need to replace all hardcoded colors with themed one, but this should probably be done as a separate task since it may require introducing new colors in existing themes (e.g. we still need to use light text colors even in light themes).
* **Other**: I reused Theme Preview screen as custom color picker screen.


ArmeF97:
He does not accept pr

PR #1540, opened 2019-11-11
Update settings.gradle

ROOTBABU:
updated the syntax

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,1 +1,1 @@
-include ':TMessagesProj'
+include :'TMessagesProj'


settings.gradle
```diff
@@ -1 +1 @@
-include ':TMessagesProj'
+include :'TMessagesProj'
```
Rch22: startup failed:
settings file 'C:\Telegram-master\settings.gradle': 1: Statement labels may not be used in build scripts.
In case you tried to configure a property named 'include', replace ':' with '=' or ' ', otherwise it will not have the desired effect.
 @ line 1, column 10.
   include :'TMessagesProj'

PR #1530, opened 2019-09-05
New pulling animation for Archive

markppv:

PR #1541, opened 2019-11-12
improved accessibility of telegram for blind users

AlexandrKozlovskiy:
In oficial telegram for android it exists problem,when it impossible accept or decline calls with talkback,if we have android seven or lower,I.E when blind users do double tap,it nothing happends,because we not get action  AccessibilityNodeInfo.ACTION_CLICK in method of custom view performAccessibilityAction and in method onTouchEvent of custom view we should be use super version of this method,I.E super.onTouchEvent and we should set clickable true in constructor of class CallSwipeView. Also in telegram,when someone busy and decline our call,instead of contentDescription with CallAgain string name we have contentDescription with VoipEndCall string name,so blind users consider,that after pressing of this button window of call in telegram probably will be closed,I.E as it works,when user press cancel button,but instead of this call start again,so it should be fixed,because it misleadin of blind users of telegram. Also in this pull request was fixed error 87,which happends,for example,if we compile telegram in android studio in windows.

### DIFF
--- a/TMessagesProj/build.gradle
+++ b/TMessagesProj/build.gradle
@@ -283,7 +283,7 @@ android {
         }
     }
 
-    defaultConfig.versionCode = 1828
+    defaultConfig.versionCode = 1829
 
     applicationVariants.all { variant ->
         variant.outputs.all { output ->
--- a/TMessagesProj/jni/libtgvoip/os/posix/NetworkSocketPosix.cpp
+++ b/TMessagesProj/jni/libtgvoip/os/posix/NetworkSocketPosix.cpp
@@ -506,110 +506,110 @@ void NetworkSocketPosix::SetTimeouts(int sendTimeout, int recvTimeout){
 	setsockopt(fd, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(timeout));
 }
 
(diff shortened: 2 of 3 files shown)


denis-skripnik:
Yes, it is necessary.
Thanks in advance for accepting Pull request.

shm-vadim:
It is very importent for screanreader users of telegram. I am waiting for merging this pull request very much!

AleksandrRozhkov:
I join for pull request. It is very importent for blind users!

ArmeF97:
@drklo does not accept pull request

AlexandrKozlovskiy:
@ArmeF97 why?

ArmeF97:
@AlexandrKozlovskiy why do you ask me?

Do you understand that this is a mirrr repo? He does not care less of what we think, issues are not even open.

He simply push sometimes his code here, just to say that telegram is opensource.

If you spend your time to push a pull request here, you are wasting your time, it will never be merged.

Thisuu:
> @AlexandrKozlovskiy why do you ask me?
> 
> Do you understand that this is a mirror repo? He does not care less of what we think, issues are not even open.
> 
> He simply push sometimes his code here, just to say that telegram is opensource.
> 
> If you spend your time to push a pull request here, you are wasting your time, it will never be merged.

So how did these features got implemented later? https://github.com/DrKLO/Telegram/pull/1530 🤔

ArmeF97:
You see it branched? I don't think so. Maybe he just browse here sometimes to copy pr's code

PR #1545, opened 2020-01-07
Fix apkdiff not checking entire files

janka102:
I noticed that `apkdiff.py` was only checking the first 4096 bytes of each file. I have added a fix so now it checks the entire contents of the file.

Before, I was able to modify the end of a file in an apk and it output that it was the same as the original. Now it correctly outputs they are different. See screenshots

photo_2020-01-06_23-24-26
photo_2020-01-06_23-24-10


DrKLO:
Thanks

PR #1550, opened 2020-03-06
Invert group chat notification order

bingfeizh:
Hi DrKLO, I really like your work and I think it is a miracle. I know you will not commit my request, however I made it. I changed the order of the group chat notification to make it more like other messenger applications and the IOS version, which I believe most users are used to. Hope you will consider about that any day.


one909:
ครับผม  😁

PR #1569, opened 2020-08-09
Notification shortcut update

alabiaga:
Remove notification shortcuts after notification is sent as per 3rd item under https://github.com/android/user-interface-samples/tree/master/People#shortcuts, to allow shortcuts to be cached.

Q: Will my shortcuts appear in the long press app launcher context menu?
A: They can depending on their rank, but if you prefer not having shortcuts appear on launcher you can remove the shortcut with ShortcutManager#removeDynamicShortcuts() or #removeAllDynamicShortcuts() after sending the notification. You can also rank other app shortcuts with higher ranking, so only those dynamic shortcuts appear on launcher.

Thanks for all your great work on this. I will push for making the API better for this. Thanks again and note that this pull request can be a throwaway and I just wanted to show at what point the shortcuts can be removed. I checked out the codebase, made these changes and was able to get this working on a flashed Android 11 device. Thanks again.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/NotificationsController.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/NotificationsController.java
@@ -2373,10 +2373,14 @@ public class NotificationsController extends BaseController {
             }
         });
     }
+    
+    private boolean unsupportedNotificationShortcut() {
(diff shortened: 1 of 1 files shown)


alabiaga:
cc\ @DrKLO

alabiaga:
Thanks @DrKLO

PR #1575, opened 2020-08-30
Upgrade MAX_ACCOUNT_COUNT

aerdnar:
I'm not sure if there are conflicts with this but MAX_ACCOUNT_COUNT should be high enough to not be a limit.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/UserConfig.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/UserConfig.java
@@ -23,7 +23,7 @@ import java.util.Arrays;
 public class UserConfig extends BaseController {
 
     public static int selectedAccount;
-    public final static int MAX_ACCOUNT_COUNT = 3;
+    public final static int MAX_ACCOUNT_COUNT = 15;
 
     private final Object sync = new Object();
     private boolean configLoaded;


alissonlauffer:
Telegram will not (probably) increase the limit of simultaneous accounts, because:
 - Above 3 accounts, we can imagine that the user would probably use them for some kind of spam or attack.
 - The app currently takes longer to open if the account limit is increased too much, since the databases for all the accounts are created/loaded at startup.
 - The app needs a rework to have multiple accounts modularly (Like TgX), without slowing down the opening speed for users which have just 1-3 accounts.

If you really need a higher limit, consider making your own builds.

Your patch to increase the account limit will not work, because to increase the limit you need to modify three different locations:
https://github.com/DrKLO/Telegram/blob/release-7.0.1_2065/TMessagesProj/src/main/java/org/telegram/messenger/UserConfig.java#L26
https://github.com/DrKLO/Telegram/blob/release-7.0.1_2065/TMessagesProj/jni/tgnet/Defines.h#L30
https://github.com/DrKLO/Telegram/blob/release-7.0.1_2065/TMessagesProj/jni/tgnet/ConnectionsManager.cpp#L137-L146

aerdnar:
Thanks for your answer! My pull request is more like a feature request because other clients can manage more than three account.
- Spam and attack are made by scripts, that's not the problem. 
- You are right but nothing changes for 1-3 accounts.
- TgX can manage multiple accounts but it seems to be closed source and it requires months to get new features. 

Own builds require to be updated and compiled at every release that not sound good. 

This function surprised me and I'm wondering if there are other parts based on max 3 accounts limit
https://github.com/DrKLO/Telegram/blob/release-7.0.1_2065/TMessagesProj/jni/tgnet/ConnectionsManager.cpp#L137-L146

dkaraush:
Increased! As a premium feature 🕺

PR #1578, opened 2020-09-10
Use AndroidX Core methods.

Isira-Seneviratne:
* Make use of methods provided in the AndroidX Core library, such as [`ContextCompat.getSystemService()`](https://developer.android.com/reference/androidx/core/content/ContextCompat#getSystemService(android.content.Context,%20java.lang.Class%3CT%3E)) and [`LinkifyCompat.addLinks()`](https://developer.android.com/reference/androidx/core/text/util/LinkifyCompat#addLinks(android.text.Spannable,%20int)) to take advantage of their backward compatibility mechanisms and backported improvements.
* Use the [`WindowInsetsCompat`](https://developer.android.com/reference/androidx/core/view/WindowInsetsCompat), [`DisplayCutoutCompat`](https://developer.android.com/reference/androidx/core/view/DisplayCutoutCompat) and AndroidX's [`Insets`](https://developer.android.com/reference/androidx/core/graphics/Insets) classes to reduce the API level checks needed.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/AndroidUtilities.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/AndroidUtilities.java
@@ -28,7 +28,6 @@ import android.content.pm.PackageManager;
 import android.content.pm.Signature;
 import android.content.res.AssetFileDescriptor;
 import android.content.res.Configuration;
-import android.database.ContentObserver;
 import android.database.Cursor;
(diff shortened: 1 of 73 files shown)


ArmeF97:
Maybe you don't understand that @DrKLO  does not accept pr

Isira-Seneviratne:
> Maybe you don't understand that @DrKLO does not accept pr

Thanks for the information.

PR #1584, opened 2020-10-15
Raise gif maximum size to align with telegram desktop

joshcoales:
It seems that the maximum gif size on telegram desktop is 1440px. Larger than that, and it displays as a video.
On iOS there is no limit to resolution, with video files up to 5000px still displaying as a gif provided there's no audio track.
This change would bring the android app into line with telegram desktop, but I'm not sure whether we could scrap the resolution check, and be more inline with the iOS app?


megapro17:
Merge it

PR #1588, opened 2020-12-01
[FIX] Add directshare shorctus in the right order

Soudini:
Currently the direct share feature is not working properly. 

The app is gathering the ten top Peers and creating the shortcuts to send them messages but they are added in the wrong order. As a result the one shown by android are not the first ones but the last ones which is not right.

The fix is trivial and my implementation is quite straightforward but as I miss any experience in android development it might not be the right one, feel free to correct me.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MediaDataController.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MediaDataController.java
@@ -2864,6 +2864,8 @@ public class MediaDataController extends BaseController {
                 }
             }
         }
+
+        Collections.reverse(hintsFinal);
         Utilities.globalQueue.postRunnable(() -> {
             try {
                 if (SharedConfig.directShareHash == null) {


Catzoy:
`Collections.reverse` adds  some computational overhead to creating a proper `hintsFinal` which can be removed by modifying the creating loop itself like so:

`final int initialHintsFinalSize = maxShortcuts - 2;`
`final ArrayList<String> hintsFinal = new ArrayList<>(initialHintsFinalSize);`
`final int indexOfLastHint = hints.size() - initialHintsFinalSize;`
`for (int i = hints.size() - 1; i >= indexOfLastHint; i--) { hintsFinal.add(hints.get(i)); }`

TheBestPessimist:
Bump. It would be really nice to have Direct Share work. 

On my Asus Zenfone 10 running Android 15 either no contacts are displayed whatsoever, or some contacts/groups which i spoke years ago are displayed. Telegram **NEVER** displayed my most recent 1-3 chats.

PR #1590, opened 2020-12-27
Delete Application.mk

priamm:


muslimlif:
سوف نبقى هنا

expoli:
Hi, sir.

I found that it seems telegram's proxy settings can't use ipv6 address?

It seems that telegram cannot use the socks5 proxy deployed on the ipv6 network, and the Windows system can be used normally.

No matter I use domain name or IP address, there is no way to complete the connection.

PR #1591, opened 2021-01-13
Call super.onBackPressed in BubbleActivity to preserve bubble

alabiaga:
Expanded bubble activity disappears onBackPressed as opposed to being collapsed into a bubble on the screen. Call super to preserve bubble.

Note that although this change fixes this UX issue, it might not be the solution you want, given the app does its own management of view transitions and task mgmt.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/BubbleActivity.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/BubbleActivity.java
@@ -299,7 +299,6 @@ public class BubbleActivity extends Activity implements ActionBarLayout.ActionBa
 
     @Override
     public void onBackPressed() {
-        super.onBackPressed();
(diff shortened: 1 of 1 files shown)


alabiaga:
cc\ @DrKLO

TMessagesProj/src/main/java/org/telegram/ui/BubbleActivity.java
```diff
@@ -309,7 +308,7 @@ public void onBackPressed() {
         } else if (drawerLayoutContainer.isDrawerOpened()) {
             drawerLayoutContainer.closeDrawer(false);
         } else {
-            actionBarLayout.onBackPressed();
+            super.onBackPressed()
```
ctwoon, line 311: you forgot ';'.

alabiaga:
closing as similar fix was applied by @DrKLO . Thanks

PR #1593, opened 2021-01-19
Add flags

loghmanAziznjad:
#Add the flag of Kurdistan
Please add the emoji of the flag of Kurdistan, this will increase the number of Kurdish language users to Telegram and make everyone happy.
Thanks to Telegram


Lwhoam:
Add automatic SMS receiving

PR #1605, opened 2021-02-14
Telegram fixes Contest

roma-sck:
fix edited message preview text in reply https://bugs.telegram.org/c/179
fix possibility to copy part of text in chat with keyboard open https://bugs.telegram.org/c/843
fix poll displaying issue after screen rotation https://bugs.telegram.org/c/676


dkaraush:
I believe it was merged, as all these bugs marked as fixed. Thanks for contribution!

PR #1606, opened 2021-02-14
Develop improvements

R12rus:
https://bugs.telegram.org/c/179
https://bugs.telegram.org/c/939
https://bugs.telegram.org/c/859


dkaraush:
Checked those bugs on the platform and they seem all fixed long time ago? Let me know if they are not.

PR #1607, opened 2021-02-14
CameraX Android Contest 2021

d-w-s-h:
1. Using CameraX in the attachment menu and for recording video messages when api ≥ 21, otherwise using legacy camera
2. Using max photo resolution available on device with target aspect 16:9, target video resolution increased to 1080p or max available on device
3. Fixed bug with cameraview glitch when photo attachment layout scrolls and cameraview goes off the screen, added cameraview parallax
4. Fixed mirroring video when recording with front face camera in attachments
5. Added exposure compensation feature in cameraview to expand the possibilities of shooting in contrasting light, added EV slider, zoom slider moved to a more intuitive location. Added 2D gesture recognition while recording to adjust zoom and EV at the same time
6.  "Use CameraX API"/"Use legacy camera API" in debug menu


Laky-64:
this code however has problems with the zoom, when you start recording a video, it starts with the wide angle zoom

PR #1612, opened 2021-02-14
Inline query for inline_geo bots not working when GPS disabled.

koloboid:
Eliminate unnecessary check, fixes https://bugs.telegram.org/c/2227


Nirzak (commented):
Hi tried mutiple geo location based bots and every one of them has requested my permission to share my location when my location is turned off. So, Can you please mention a bot in which you are facing such issue. May be I can test it with that bot.

koloboid:
@Nirzak  Oh my, missed your comment, sorry. As I see, this issue was fixed with last update of Telegram for Android (so, to reproduce, you need to install apk version that is current for February'21)

Right solution is now in repo: https://git.io/JOgtw

Going to close bug and PR :|

PR #1624, opened 2021-03-21
fixed country code and phone no. population logic

xrafi:
**Problem** 

When people try to log in or sign up, the characters in the phone number are used to set the country code, which is incorrect.
This results in country code and phone number being populated incorrectly. 

Example of incorrect behaviour :  
Phone number below is : 6692049100
Country is US

However, 66 is taken for country

image

**Solution**
Use getSimCountryIso() to retrieve country codes

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/LoginActivity.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/LoginActivity.java
@@ -1711,30 +1711,15 @@ public class LoginActivity extends BaseFragment {
                     }
                     if (!newAccount && allowCall) {
                         String number = PhoneFormat.stripExceptNumbers(tm.getLine1Number());
-                        String textToSet = null;
-                        boolean ok = false;
+                        String countryCode = tm.getSimCountryIso();
                         if (!TextUtils.isEmpty(number)) {
-                            if (number.length() > 4) {
-                                for (int a = 4; a >= 1; a--) {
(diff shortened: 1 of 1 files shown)


rafismugmug:
@DrKLO

Luxxy-787:
Try manually

Mahmed21 (changes requested):
Leavae

Nirzak (commented):
Well, I didn't get your point. I am using the latest Telegram app on my mobile and I always got the correct country code by selecting country. Here on your image, you have selected Thailand, and +66 is the correct code for Thailand. So, what's the problem? I tried it just a few moments ago. And it also gives +1 as a country code when selected USA as a country. So I think it's working just fine.

rafismugmug:
When you come to login screen, The country code should be automatically populated from the sim card or phone number. Here country code should have been +1, but in the screenshot, it's +66 Thailand, which is incorrect. You don't need to manually select that. The PR fixes the country code population logic, so that correct country code is detected from the phone.

Nirzak:
> When you come to login screen, The country code should be automatically populated from the sim card or phone number. Here country code should have been +1, but in the screenshot, it's +66 Thailand, which is incorrect. You don't need to manually select that. The PR fixes the country code population logic, so that correct country code is detected from the phone.

Yeah my bad actually. Now I see.

Nirzak (approved):

PR #1625, opened 2021-03-29
Fix incorrect unread messages counter

IAmKapuze:
Table "messages" in cache4.db doesn't keep all unread messages. If, for example, any channel have 100 unread messages, table "messages" may content only one (then COUNT(mid) will be = 1). Because of this, counter may display an incorrect value after update.
I found the "still_unread_count" field in TL_updateReadChannelInbox and TL_updateReadHistoryInbox, the use of which solves the problem, but I don't know why it's not used.
Maybe there is a preferred way to solve the problem?

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MessagesController.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MessagesController.java
@@ -11670,16 +11670,22 @@ public class MessagesController extends BaseController implements NotificationCe
             } else if (baseUpdate instanceof TLRPC.TL_updateReadHistoryInbox) {
                 TLRPC.TL_updateReadHistoryInbox update = (TLRPC.TL_updateReadHistoryInbox) baseUpdate;
                 long dialog_id;
+                long message_id = update.max_id;
+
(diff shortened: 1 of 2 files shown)


IAmKapuze:
Added a description (with video) of the problem at the link:
https://bugs.telegram.org/c/4849

nizomovasadbek:
Hello I have an idea

Yak, 2-may, 2021, 19:27 IAmKapuze ***@***.***>:

22072012 (approved):

Playez:
You should ping @xaxtix , drklo doesnt work on telegram android client rn,
xaxtix does it

вт, 11 янв. 2022 г., 6:28 PM Alexandr Keyp ***@***.***>:

IAmKapuze:
@xaxtix, tdlib use still_unread_count variable to update messages counter. May be, this code can be taken as a basis ?
https://github.com/tdlib/td/blob/master/td/telegram/MessagesManager.cpp#L6598

IAmKapuze:
Fixed in 8.9.0

PR #1630, opened 2021-06-01
Support camera trigger from headset / selfie-stick

superatrain:
Adds keycodes to shutter control to enable use with wired remotes / selfie-sticks.


dkaraush:
I see this PR was merged, thanks!

PR #1636, opened 2021-07-07
fix bug: can't accept incoming call when we don't have permission

rezaei7:
fix bug: can't accept incoming call when we don't have record audio or camera permission


rezaei7:
@DrKLO

Digital-Lemon:
@rezaei7 Just planned to look into this and I see there's already a PR open! Thanks for doing this!

I don't do Java, so I after some glancing, I think this fixes the same issue I have. I grant the microphone permission `ONLY THIS TIME`, which I do for every single app that requires microphone access.

With Telegram app, for outgoing calls, I always get the permission prompt, which I accept and the call takes place.
For incoming calls however, permission prompt doesn't get displayed, so when I answer the call, if Android removed the permission since I had a call "last time", Telegram won't get any mic input, and the other party will not hear me. In those cases, I need to hang up, and call them back myself, which will trigger the permission prompt.

@DrKLO Could we have this merged for the next release please?

dkaraush:
I see this PR was merged, closing.

PR #1640, opened 2021-08-09
improved accessibility of telegram

AlexandrKozlovskiy:
In this pull request i want to improve accessibility of telegram. First of all,i optimize information about forwarded message,and fixed problem with virtualViewId. Also i added contentDescription to bot menue view,and added expand/collapse accessibility action,to blind users can know,whether this view expanded/collapsed. Also i fixed problem,when in acceptDeclineView we have sound from talkback,which shows,what there is no next/previous node on the screen. Also i fixed possible problems,when blind users focus on some virtualview node and during explorebyTouch they focused on this node again. Also i added seekbar not only for voice messages,but for video messages too. And probably other fixes,which i not mentioned.

# known issues,which was before modifying of sources of telegram:

1. Please record video message. After it you can't play it before sending,because we haven't button for this,as for voice message.

I dont know,whether this possible for sighted people,but if no,it should be implemented for video messages too.

2. Please focus on some video message. You will hear usefull information,including duration of this message,e.g 12 seconds. Please press double tab and focus on it again. Duration will be 11 seconds. So,what information is more correct,before playing,or after starting of the playing. When we close this chat,and open it again,duration of the message not changed,but if we open this chat after very long time,problem repeat again. and this for each video (only video,not voice,message).


AlexandrKozlovskiy:
@DrKLO I also added a feature to display,whether contact is mutual. I dont know,how to add this line to telegram translations,so i simple added string value via code.

PR #1642, opened 2021-08-26
fix recent searches

ivk1800:
search queries never displayed together in recent searches:
phone number and peer-to-peer is calls
save to gallery and distance units


dkaraush:
Already merged, thanks for PR!

PR #1670, opened 2021-12-27
Added missing permission to work on Android 11

Mirivan:


singhalrishi27:
> _No description provided._

Nope you can't use manage external storage on Android 11 without a good reason only some file manager apps can use this permission

YaCodesDevelopment:
You could, at least on standalone version of the app (org.telegram.messanger.web), so it should be added to standalone version.

Mirivan:
> > _No description provided._
> 
> Nope you can't use manage external storage on Android 11 without a good reason only some file manager apps can use this permission

lm.

PR #1675, opened 2022-01-29
Revising all texts, adding consistency

mcsaeid:
Consistency is key.

I would be more than happy to provide reasoning for each change made.


mcsaeid:
@DrKLO @xaxtix

PR #1684, opened 2022-02-17
Fix "Send Message" & "Send without sound" width from ShareAlert.java

Laky-64:
I took the trouble to solve this little graphic bug, since the parent layout need createFram, but we use createLinear as parameters, this means that the button does not expand completely, so to make sure to keep the buttons that are below the other and at the same time using the createFrame, I directly created a parent linearlayout (Using as parameters createFrame) and then add the various buttons as children keeping the createLinear, in this way the father will take the maximum length thanks to createFrame and the children will follow it


ghost (approved):
Nice

PR #1685, opened 2022-03-05
Fix crash

Nightprince:


akirataguchi115 (changes requested):
Which crash/bug does this fix again?

Nightprince:
> Which crash/bug does this fix again?

Fixed with this commit :

akirataguchi115:
Good to hear. Maybe close this one then?

PR #1687, opened 2022-03-07
space waste fix

nikitasius:
for large tablets in landscape orientation
no need to pickiup the small one, just use the x

before
space_waste_nopr
.
after
space_waste_pr


dkaraush:
The longer the line, the harder it is to read text jumping with your eyes from line to line.
Although it can be made slightly wider, but definitely not the full width of a screen.

If you want to use the full width of a screen, use code block in formatting.

PR #1696, opened 2022-04-08
Optimized PNGs

VadimMuhtarov:


VadimMuhtarov:
Optimized PNGs via oxipng to reduce footprint

PR #1697, opened 2022-04-10
Fix switch_inline_query_current_chat

prtngn:
Then bot send keyboard with switch_inline_query_current_chat param, app wrong select user id and set Field text to @user_name and query, but right is @bot_user_name and query.


vaishnavi-ramanujapuram:
HI, is this issue still open? if not could you elaborate, as i would like to contribute.

PR #1644, opened 2021-09-02
Improved accessibility of telegram and quality of voice messages

AlexandrKozlovskiy:
In this pull request i want to improve accessibility of telegram. First of all,i optimize information about forwarded message,and fixed problem with virtualViewId,if message is forwarded. Also i added contentDescription to bot menue view,and added expand/collapse accessibility action,to blind users can know,whether this view expanded/collapsed. Also i fixed problem,when in acceptDeclineView we have sound from talkback,which shows,what there is no next/previous node on the screen. Also i added seekbar not only for voice messages,but for video messages too. Also i added improvement of quality of recording of voice for voice and video messages.

And probably other fixes,which i not mentioned.

# known issues,which was before modifying of sources of telegram:

1. Please record video message. After it you can't play it before sending,because we haven't button for this,as for voice message.

I dont know,whether this possible for sighted people,but if no,it should be implemented for video messages too.

2. Please focus on some video message. You will hear usefull information,including duration of this message,e.g 12 seconds. Please press double tab and focus on it again. Duration will be 11 seconds. So,what information is more correct,before playing,or after starting of the playing. When we close this chat,and open it again,duration of the message not changed,but if we open this chat after very long time,problem repeat again. and this for each video message (only video,not voice,message).

PR #1704, opened 2022-05-05
Fix voice message sample rate (again)

sermart1234:
https://github.com/DrKLO/Telegram/pull/1566
But voice messages still use 16kHz. I set 48 kHz.


egormanga (approved):

itsreallyv1eds:
in almost all modified clients this is done, and they even do it in the first place. I think if this is added to the official client, it will not become superfluous, especially since there are no complaints

itsreallyv1eds (approved):
in almost all modified clients this is done, and they even do it in the first place. I think if this is added to the official client, it will not become superfluous, especially since there are no complaints

Attorelle:
@AlexandrKozlovskiy previously suggested this, with better voice processing in #1644 :)

Playez (approved):
### 

- [x] 1. > [`_****_`](url)

dkaraush:
I see this was merged, thanks for the PR!

PR #1709, opened 2022-06-19
Fix for a strange vibro flow in telegram

nikitasius:
The original flow:

- check `hasEmojiSupportVibration`
- if not, then vibro
- WTF?!

so fixed the way "if emoji support vibration we will activate the vibro".
If i'm wrong, please explain me your idea :smile:


dkaraush:
if it has support, it will vibrate in animation. such emojis as hearts have their own vibrations when sent

PR #1710, opened 2022-06-21
Fixed a Duplicate implementation from the build.gradle

Laky-64:
I found a repeat of the com.android.billingclient implementation


dkaraush:
Outdated.

PR #1711, opened 2022-06-25
Fix typo in strings.xml

realdeveloperongithub:
Typo


dkaraush:
Thanks for PR, these kinds of typos are better to be fixed in https://translations.telegram.org/

https://translations.telegram.org/en/android/unsorted/PsaInfo_covid

PR #1714, opened 2022-07-09
Fixed Theme Parser

Laky-64:
When there are a space on the theme, this will cause some problems with the parser


Nightprince:
Confirmed, but we also have in:
ThemeCell.java
ThemesHorizontalListCell.java
ChatThemeBottomSheet.java

umarufum:
Write the technical statement of the platform

PR #1717, opened 2022-07-16
Change criteria for high perfomance class

E1ecti:
Modern devices with a given frequency can be defined as high-performance


YaCodesDevelopment:
> Modern devices with a given (2.0 GHz) frequency can be defined as high-performance

I don't think so, 2.0 GHz for now is kinda not enough powerful to be defined as such. Most High-end devices for the last 5 years got over 2.5 GHz-clocked CPUs.

YaCodesDevelopment:
And BTW you've changed definition of "Mid-range" class, not "High performance" one

PR #1720, opened 2022-08-21
Initial watch support

Yakov5776:
**When on watch:**
 - Enter-to-send will be enabled by default
 - Font size will be set to 12 by default
 - Pinned messages banner will be hidden by default (due to lack of space on screen)


Someone3731:
Hello,

I would like to know if the following teams are still active : TSF, abuse, Telegram Auditions and moderator team (@username_bot).

I have contact each team for more than a month ago and several times but I never had any answer.
I know I'm not the only one in this situation and I think they are understaffed but why are they not doing anything? A lot of people are waiting for Telegram Auditions to answer and be able to help.

Note: I know this is the place but I'm desperate because they don't answer anywere.

Thanks in advance to people who will give me more informations about the situation.

Yakov5776:
Closed due to inactivity.

PR #1721, opened 2022-08-29
auto translation feature in chatActivity no need to translate individual messages 

rajashish140:
auto translation feature is added in chatActivity.java file
 auto translation function is called on recycler view scroll state == idle 
translation button is added 
 parsing of #hashtags, @usernames and url is done
Can't parse bold and texturl as they are subject to translation although we are trying to parse with string matching if possible


dkaraush:
Such API is no longer used, all entities are now supported by translating as HTML, including even animated emojis. Great PR though!

PR #1722, opened 2022-08-29
Fix return value of onKeyDispatch instead of returning always false.

mbarashkov:
onKeyDispatchEvent method in LaunchActivity always returned false if event was bubbled up to super.dispatchKeyEvent.
This created various bugs with hardware keyboards: for example, backspace button was handled twice, while cursor buttons didn't work at all.
This pull request fixes those issues.


dkaraush:
Oh, yes! I remember this bug, took a while to find the cause. Thanks for the PR!

PR #1594, opened 2021-01-30
Handle keyboard and D-PAD selection in lists

mbarashkov:
This pull request enables selection of list items in RecyclerListView by pressing Enter or Dpad's Center button (or other keys as specified by https://developer.android.com/training/tv/start/controllers) on list items. This works on lists where item's focusable property is set to true, for example, on the applications' main chats list.

PR #1723, opened 2022-08-29
Fix keyboard dpad navigation in lists

mbarashkov:
This pull request enables selection of list items in RecyclerListView by pressing Enter or Dpad's Center button (or other keys as specified by https://developer.android.com/training/tv/start/controllers) on list items. This works on lists where item's focusable property is set to true, for example, on the applications' main chats list.
This replaces my similar pull request https://github.com/DrKLO/Telegram/pull/1594 with updated code.


ljambdadelta:
Dear Telegram team,

I would like to kindly ask you to consider merging this PR.
Without this fix, a number of touch-screen-less devices (keypad / D-pad based) are unable to properly operate Telegram, as the Enter / DPAD Center key does not activate the currently selected dialog in the main chat list.

This patch restores expected keyboard navigation behaviour by mapping the select key to the existing item click path, making Telegram usable on such devices again.

Thank you for your time and for maintaining Telegram Android.

PR #1728, opened 2022-11-08
Fixed inline bots not working correctly

Laky-64:


dkaraush:
Thanks, I see it was merged already.

PR #1729, opened 2022-11-08
Fixed two memory leaks

Laky-64:


dkaraush:
I see those changes were already merged, thanks for the PR!

PR #1731, opened 2022-11-13
Fixed incorrect order in proxy list

arsLan4k1390:
Recently added proxies will appear on the top instead of bottom of the list of proxies (https://bugs.telegram.org/c/1128)


dkaraush:
I see the code is merged, thanks for PR!

PR #1732, opened 2022-11-22
Fixed app not responding after creating chat preview

arsLan4k1390:
We have to use _newLastFragment.onTransitionAnimationEnd(true, true);_  inside _onCloseAnimationEnd_


dkaraush:
sorry, but LNavigation was removed from the app, closing

PR #1738, opened 2022-12-29
Fix issue with search in ActionBar

Laky-64:
This pull request fixes an issue where searching for the "from:" keyword in the ActionBar would cause the app to crash. The issue was caused by attempting to access a line in the layout that did not exist. The solution is to add a check to ensure that the line offset is within the bounds of the layout before attempting to access it. This fix will prevent the app from crashing when searching for "from:" in the ActionBar.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/Components/EditTextEffects.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/Components/EditTextEffects.java
@@ -147,6 +147,7 @@ public class EditTextEffects extends EditText {
 
             Layout layout = getLayout();
             if (text instanceof Spannable && layout != null) {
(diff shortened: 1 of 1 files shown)


TMessagesProj/src/main/java/org/telegram/ui/Components/EditTextEffects.java
```diff
@@ -147,6 +147,7 @@ protected void onTextChanged(CharSequence text, int start, int lengthBefore, int
 
             Layout layout = getLayout();
             if (text instanceof Spannable && layout != null) {
+                if (start > layout.getLineCount()) return;
```
dkaraush: start is an offset, not line index; did you mean layout.getText().length()?

PR #1739, opened 2023-01-11
Update AndroidManifest.xml

aria1991:
1- I've added <uses-permission android:name="android.permission.INTERNET" /> and other permission that should be in the app, also added activity to launch the app and activity to handle the passport.

2- It's better to generate a new API key for the Google Maps feature instead of using the one provided in the script. This will ensure the security of the application and prevent unauthorized access to the Google Maps feature.

3- The uses-permission and uses-feature elements should be carefully reviewed and only the necessary permissions should be requested. It is important to be transparent with the users about the permissions that the application needs and the reasons why they are necessary.

4- It's better to use android:protectionLevel="signature" for permission that are not critical to the app operation and do not contain sensitive information and if possible, implement runtime permissions which allow users to grant permissions at the time the feature is used rather than at the time of installation.


TaiTaiTai777:
https://youtu.be/uAJRd8BI804?si=BqOCfgGhz6XRsgJF

PR #1747, opened 2023-03-06
Remove WaveDrawable completely.

0x7FFF:
WaveDrawable is unused as a class, so there's no need for it.


SpangleLabs:
It is generally useful to have constants defined in one place, such that they can be edited once and changed everywhere at the same time.
Using "actual numbers" around the place like you have done is often called "magic numbers" because it is suddenly unclear what the numbers mean, or where they came from.
The java compiler will likely do the math once, and use that value, as it knows at compile time that these constants are constants, and won't need recomputing every time.

Richardsom41:
Gracias

PR #1748, opened 2023-03-07
Remove redundant "if (BuildVars.LOGS_ENABLE)" checks.

0x7FFF:
FileLog.# contains this check inside of it's body, so it is a good idea to delete it.


tehcneko:
There is a thing called compiler optimize...

rezaei7:
don't to this , because when LOGS_ENABLED = false , string argument moved to StringPool and this cause of Out of memory

saeedzada:
https://github.com/DrKLO/Telegram/pull/1748#issuecomment-1458337672

PR #1750, opened 2023-03-16
Native lottie optimizations.

0x7FFF:


Hassan222-ux:
0x7FFF:optimize_lottie

PR #1751, opened 2023-03-18
feat: mobile coin integration proposal

baermathias:
Dear Maintainers,

before creating a PR with code I wanted to check whether this proposal would be approved at all.

This idea is to integrate [Mobile Coin](https://mobilecoin.com/) into Telegram. As you may already know, Signal messenger recently added [support for Mobile Coin](https://www.wired.com/story/signal-mobilecoin-cryptocurrency-payments/), and it has been widely hailed as a significant step forward for both privacy and cryptocurrency adoption.

Mobile Coin is a lightweight cryptocurrency that enables private, fast, and efficient transactions on mobile devices. With Mobile Coin, users can send and receive payments without revealing their identity or personal information. This makes it an ideal fit for Telegram, which is known for its focus on privacy and security.

Here are a few reasons why I believe adding Mobile Coin support to the Telegram would be a great idea:

Increased Privacy: By integrating Mobile Coin, Telegram users would have access to a more private and secure payment system. Mobile Coin uses zero-knowledge proofs to ensure that transactions are completely anonymous, which would be a significant improvement over existing payment options.

Faster Transactions: Mobile Coin transactions are fast and lightweight, making them ideal for mobile devices. This would be a significant improvement over existing payment options, which can be slow and clunky on mobile devices.

Positive Media Coverage: Adding Mobile Coin support to the Telegram repository would likely attract media attention and generate positive coverage for both projects. This could lead to increased adoption and more users for both projects.

I believe that integrating Mobile Coin into the Telegram would be a significant step forward for both projects and would help to further cement Telegram's reputation as a leader in privacy and security. I look forward to hearing your thoughts on this proposal and hope that you will consider it seriously.

Here are also two images of how this could look:

Step 1:
step1_mobile_coin

Step 2:
step2_mobile_coin


playez:
ну и нахуя когда есть боты по типу @wallet и @cryptobot

baermathias:
@playez What is the argument against the use of ChatGPT? It served its purpose and created a PR text and saved me valuable life time. 
If you have constructive criticism to bring against the feature itself, please do.

PR #1755, opened 2023-03-28
Performance Improvement: Message Translation

philippelaporteproduct:
I'm a developer at productscience.ai, we develop AI-powered instrumentation and profiling tools for mobile. We're big fans of Telegram. Our team is multi-lingual and we were super excited when Telegram released the auto-translation feature. However, we noticed that it was a bit slow, so I decided to investigate further using our toolset.

Turns out there's a hardcoded delay of 200ms that postpones the network request being sent. This is a snippet from the code: 

```
private static final int GROUPING_TRANSLATIONS_TIMEOUT = 200;
...
AndroidUtilities.runOnUIThread(pendingTranslation.runnable, GROUPING_TRANSLATIONS_TIMEOUT);
```

Sometimes these kinds of delays can slip through the cracks even when they're no longer necessary. With Product Science tool I was able to try out a fix and test it right away. So, I went ahead and rescheduled the network request for TL_messages_translateText earlier, and the fix reduced the wait time for users on mid-end devices like Galaxy A52 by around 17%. I did a bunch of tests and the fix doesn't seem to mess with the feature's functionality.

Would appreciate your feedback

Trace Before Fix:

Screenshot 2023-03-28 at 11 14 33

Trace After Fix:

Screenshot 2023-03-28 at 11 13 35


dkaraush:
Done at 89b0e317534f614112751e0b5c0eabe8241599cd.

PR #1780, opened 2023-08-25
Fix implicit setting of the invalidate flag

FUZIK:
Fix corner animations for DialogCell
A couple of before/after screenshots

before

after


aliyevakhalida (approved):

dkaraush:
Thanks! This fix was merged into the release

PR #1782, opened 2023-09-24
Fix premium subscribers count and their percentage in channel statistics 

ayebrian:
Very easy to fix.
### Before
image

### After
image


dkaraush:
Thanks for the PR, was fixed at its time.

PR #1790, opened 2023-11-04
Allow sending pictures up to 2560px

gobonja:
- This PR will allow sending pictures up to 2560px.

- Telegram API already allows to send photos with that size.

- Telegram Desktop and some forks (e.g. Nekogram) also supports this.

- There is also bug report for this: https://bugs.telegram.org/c/15370

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/AndroidUtilities.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/AndroidUtilities.java
@@ -2477,7 +2477,7 @@ public class AndroidUtilities {
 
     public static int getPhotoSize() {
         if (photoSize == null) {
-            photoSize = 1280;
+            photoSize = 2560;
         }
         return photoSize;
     }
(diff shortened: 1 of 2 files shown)


Kris35hoog:
Ok

gobonja:
@dkaraush @DrKLO @xaxtix any news about this PR?

Zombiles:
Screenshot_2024-01-24-20-49-03-813_com vkontakte android

QuAzI:
Waiting for this feature as currently photos quality in TG is terrible and TG gallery works ugly with photos which uploaded as files

gobonja:
> This increases the initial texture buffer allocation by significant amount in painting screen. You might want to apply this increase only for photos that are already that size or larger. Considering reports of `OutOfMemoryException`, it would be also great to do that only for HIGH and AVERAGE devices...

As far as I can see, this functionality is already implemented: https://github.com/DrKLO/Telegram/blob/d494ea8cb59f6ff1af5faca9f5f651abcec618c0/TMessagesProj/src/main/java/org/telegram/messenger/MessageObject.java#L8531-L8536

It will use actual image size if size is not bigger then 2560px, if it is bigger it will downscale image to 2560px.
So it will never increase image size, just allow images up to 2560px.

fhnb16:
Still waiting for this feature and suffering with overcompressed images..

vovmomyt (approved):
Like

PR #1791, opened 2023-11-06
Fix text scaling in web-view

Svechnikov:
Telegram enforces the normal scale type in the app itself but not in web-view. This might lead to unexpected results in all web apps in general, but it's especially noticeable in Flutter web-apps.

Below you can see examples with an [app](https://gallery.flutter.dev/#/reply) from the official Flutter gallery. They were captured on an emulator (Pixel 3A API 34 extension level 7 x86_64) but we also reproduced the issue on physical devices (Samsung S23 Ultra, Google Pixel 7A API 33).

In order to fix this, we can use `WebSettings.LayoutAlgorithm.TEXT_AUTOSIZING`. This shouldn't affect/break anything else.

| System text size | Before | After |
|--------|--------|--------|
| Screenshot_20231106_221825 | Screenshot_20231106_221836 | Screenshot_20231106_222102 |
| Screenshot_20231106_221909 | Screenshot_20231106_222005 | Screenshot_20231106_222129 |


dkaraush:
Thanks, merged!

PR #1794, opened 2023-12-06
GLIconSettingsView visibility condition

vladimirlisovskij:
Hello there!

While using the app from Google Play I accidentally noticed that long pressing the star in the premium mode menu causes the settings screen to open.

This small fix should remove this behaviour for release builds


dkaraush:
Thanks, merged!

PR #1801, opened 2024-02-06
Feature/contacts enhancements

Quatern1on:
- Add multiselect to the contacts screen to present the ability to delete multiple contacts at once. The user can undo the deletion via the toast button within 5 seconds.
- Freshen up the 'Invite Friends' section by adding gradient background to the cells of contacts who are not yet registered in Telegram. The corresponding gradients were reused from the avatar builder.
- Fix a bug with the layout when searching in the 'Invite Friends' section. Reuse skeletons and stubs from 'Contacts' section.


dkaraush:
Merged!

mohamad1101:
the variable isgoodprovider in route: TMessagesProj/src/main/java/org/telegram/messenger/ContactsController.java
how get values and using for what?

PR #1802, opened 2024-02-09
Fix MiniApp scrolling: when scroll up - the webview is closing

fede4ka1245:
### **The bug**
The mini app is minimizing, when user creates these actions:  reversible scrolling, scrolling inside containers, and swiping elements down. The bug is not permenent: it happens often but not always. Additional details [here](https://bugs.telegram.org/c/36664)

### **Reproduce**
1) Open android tg app
2) Open [this tg bot](https://t.me/AndroidScrollIsueBot)
3) Open web app with menu button and expand it
4) Scroll down
5) Scroll up (here it is. Web view is closing)

Here is how code of this app look like: https://stackblitz.com/edit/typescript-gufuwl?file=index.ts

### **Importance**
This problem prevents users with Android devices from working with Mini Apps. Developers are unable to create applications with the full range of advantages of web applications: reversible scrolling, scrolling inside containers, and swiping elements down. Therefore, I propose a solution to avoid all of these issues.

### **TLDR of the commit:**
- ignore swiping the container with web app when it's expanded
- add menu's button "Minimize" to minimize web app

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/Components/BotWebViewMenuContainer.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/Components/BotWebViewMenuContainer.java
@@ -144,6 +144,8 @@ public class BotWebViewMenuContainer extends FrameLayout implements Notification
             botMenuItem.addSubItem(R.id.menu_reload_page, R.drawable.msg_retry, LocaleController.getString(R.string.BotWebViewReloadPage));
             settingsItem = botMenuItem.addSubItem(R.id.menu_settings, R.drawable.msg_settings, LocaleController.getString(R.string.BotWebViewSettings));
             settingsItem.setVisibility(View.GONE);
+
+            botMenuItem.addSubItem(R.id.menu_minimize, R.drawable.msg_go_down, LocaleController.getString(R.string.BotWebViewMinimizeView));
         }
     }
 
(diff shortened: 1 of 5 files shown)


fede4ka1245:
In additional, I can suggest adding swipe only when user touches the part of container which is higher than webview. The place where the name of webApp is written

Screenshot 2024-02-10 at 00 11 27 1

btbsb:
I have the same problem, has anyone fixed it?

fakeou:
got the same question, when it will be fixed,

dkaraush:
The fix for conflicting swipes with scrolling/touches is up now in 11.0. 

Swipe is disabled when:
- `event.preventDefault()` is run inside `touchstart` or in first `touchmove` touch event
- `scroll` event is run after `touchstart`, and target container is not at the top
- container underneath the touch changes its styles or attributes (was done primarily for js-based carousel widgets, that change `transform: translationX(?)`)
- bot runs `disableVerticalSwipes` in [Bot API](https://core.telegram.org/bots/webapps)

Such fix should, by default, stop swipes when scrolling inside inner containers in HTML.
It is recommended to do `e.preventDefault()` when implementing custom drag&drop or own swipes, to prevent app swipes on particular touches, and not disabling swipes for the whole mini app.

You can look at the source code for this implementation: https://github.com/DrKLO/Telegram/blob/3409cb1c72e50bbc2bd2eced7ba98aabc62a1367/TMessagesProj/src/main/res/raw/webview_app_ext.js

---

This fix itself came after development of in-app browser. If you wish to disable X/Y swipes in generic browser, you can also put such meta tags:
`<meta name="tg:swipes:x" content="none">`
`<meta name="tg:swipes:y" content="allow">`
This works only in in-app browser. Use `disableVerticalSwipes` in Bot API to disable swipes in mini apps.

PR #1805, opened 2024-03-08
fix: enhance support of HDR10 in video editor

Quatern1on:
- Fix issue with bleached colors of HDR10 video in video editor preview
- Fix issue with bleached colors of HDR10 video after conversion, if filter effects are applied
- Fix issue with HLG video dramatically changing exposition after conversion
- Add support to full color range in HDR10 video
- Make video editor use the same tone mapping method for both PQ and HLG when converting to SDR


harshil21:
Thank you!! Just saw this merged on 10.9.3 beta. This was a pretty big complaint of mine.

dkaraush:
Merged!

PR #1809, opened 2024-05-02
Accessibility inprovements

AlexandrKozlovskiy:
In this pr i made some accessibility improvements and fixed memory leak. Now more information should be announced. Also for now it should be announced and displayed,if someone write not only as user,but as chat or channel.


Rappid6974 (approved):

Rappid6974 (approved):

Arash63Dehghan (approved):
[]() v / 0

PR #1812, opened 2024-05-18
Fix saving files with the same names

dsatyanov:
@dkaraush Do you remember or could you help with understanding why do we need this removed code here?
For me it looks like speed boost/optimization because we don't touch FilePathDatabase in this case - is it correct?

Because of the code there is the bug in the app:
https://bugs.telegram.org/c/35162
https://bugs.telegram.org/c/32951
when there is more than one file with the same name we could "save to downloads" only the first of the cached files, because other files are saved internally with suffix " (_number_)" (see [FileLoadOperation](https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/java/org/telegram/messenger/FileLoadOperation.java#L1584))

So it seems like we **have to** get file path from [FilePathDatabase](https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/java/org/telegram/messenger/FilePathDatabase.java), where the file path is saved properly ([see putPath](https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/java/org/telegram/messenger/FilePathDatabase.java#L245)) - is it correct?

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MediaController.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MediaController.java
@@ -4369,18 +4369,18 @@ public class MediaController implements AudioManager.OnAudioFocusChangeListener,
                                 }
                             }
                             if (path == null || path.length() == 0) {
-                                path = null;
-                                TLRPC.Document document = message.getDocument();
-                                if (!TextUtils.isEmpty(FileLoader.getDocumentFileName(document)) && !(message.messageOwner instanceof TLRPC.TL_message_secret) && FileLoader.canSaveAsFile(message)) {
-                                    String filename = FileLoader.getDocumentFileName(document);
-                                    File newDir = FileLoader.getDirectory(FileLoader.MEDIA_DIR_FILES);
-                                    if (newDir != null) {
-                                        path = new File(newDir, filename).getAbsolutePath();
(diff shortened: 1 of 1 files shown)


Aokromes:
imho, if you want to stop using some code better remove it, don't comment it.

dsatyanov:
> better remove it, don't comment it

Agree 100%
It's not the final fix it's just to highlight the cause of the bug.
But yes - now I see that I could just remove the code and we'd still see it

postboy:
I've also experienced this bug. What's worse, I do believe this bug poses a security vulnerability in Telegram for Android. I've tried to contact security@telegram.org and dkaraush@gmail.com but got no response in 2 weeks so I decieded to disclosure this. Here's my writeup:
https://telegra.ph/Telegram-for-Android-Downloaded-Files-Cache-Poisoning-06-21

postboy:
Here's my writeup about this bug in Russian: https://habr.com/ru/articles/823612/

postboy:
Status of one of the bugs on Telegram bug tracker ia changed to "Fix is coming". Telegram channel Beta Info (https://t.me/betainfo) published beta build with supposedly fixed bug.

postboy:
Fixed in version 10.14.3 (4927) from Google Play.

PR #1815, opened 2024-05-24
fix:Solve the problem that the soft keyboard of webview cannot pop up after closing the opened tg_link page

longqingzhao:
https://github.com/Telegram-Mini-Apps/issues/issues/21


longqingzhao:
After testing, ios and other platforms do not have this problem.

longqingzhao:
@dkaraush @DrKLO @xaxtix  I sincerely hope that we can address the soft keyboard focus issue in the mini app as soon as possible. As our business heavily relies on the use of input boxes, and we need to share and make payments within Telegram, a prompt resolution is of utmost importance. Thank you for your attention and assistance in this matter.

dkaraush:
Fix is incoming in newer version.

PR #1818, opened 2024-07-07
Fix MarkAllAsRead for archived forums

GuillerLT:
Issue repro steps

1. Archive forum (group with topics)
2. Receive a message in any of the topics
3. Long press "Archived Chats"
4. Press "Mark all as read"
5. Observe that the forum still has unread messages

I added the same logic that is already being used in
```java
// DialogsActivity.java
private void markAsRead(long did) {
  // ...
  if (getMessagesController().isForum(did)) {
    getMessagesController().markAllTopicsAsRead(did);
  }
  // ...
}
```


GuillerLT:
https://bugs.telegram.org/c/41582

dkaraush:
Merged, thanks!

PR #1819, opened 2024-07-13
Fix the bot_id variable type

lana-codes:
Bugfix the `bot_id` variable type, changing it from `int` to `long`, because it should be long instead of int.

int max size in Android: 2147483647

But the IDs of the newly created bots are already larger than int max value.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/LaunchActivity.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/LaunchActivity.java
@@ -4640,7 +4640,7 @@ public class LaunchActivity extends BasePermissionsActivity implements INavigati
             });
             presentFragment(fragment, false, true);
         } else if (auth != null) {
-            final int bot_id = Utilities.parseInt(auth.get("bot_id"));
+            final long bot_id = Utilities.parseLong(auth.get("bot_id"));
             if (bot_id == 0) {
                 return;
             }


lana-codes:
Before the fix, Telegram Passport auth didn't work if the `bot_id` was larger than the int max size.

Examples:

working:
`tg://resolve?domain=telegrampassport&bot_id=2147483646&nonce=1&payload=nonce`

not working (in the bug-affected version):
`tg://resolve?domain=telegrampassport&bot_id=2147483648&nonce=1&payload=nonce`

After the fix, the links also work where the `bot_id` is larger than `2147483647`.

dkaraush:
Merged, thanks!

ahmedkh19:
@dkaraush This problem is still there, it's causing an issue with Telegram Passport auth for large bot IDs. Anything over 2,147,483,647 isn't working right.

Can you take a look ?

lana-codes:
@ahmedkh19, you do know that this is still only in the master branch, right? Only the pull request has been accepted, and the new version containing the fix hasn't been released yet.

I tried the fix, and it worked.

lana-codes:
I made a screen recording with Android Studio. I opened the following test link:
`tg://resolve?domain=telegrampassport&bot_id=2147483648&nonce=1&payload=nonce`

[Screen_recording_20240720_161559.webm](https://github.com/user-attachments/assets/48c87ca2-b348-403d-b124-0f4f4f57dc5c)

The result is perfect. Telegram handles the request and displays that the bot id is invalid, which is correct. This is the expected behavior because it means that it reaches the Telegram Passport request, which returns an error message.

Previously, in the bug-affected version, there was no notification or anything. Telegram just opened blankly because there was a `return` in the code when bot id is incorrect (in this case, the incorrect behavior should be interpreted as when it exceeds the maximum range of an integer, causing the bot id to become `0`).

PR #1821, opened 2024-07-24
feat: add cam2api features

cnwxu3ea:
added autofocus and stabilization for video recording


NayamAmarshe:
It would be great to merge this! Does this also enable CameraX or is that a different thing?

PR #1825, opened 2024-09-22
Ensures that the service is recognized as handling phone calls

mehebub648:
After recent changes to VoipService.java, the app stopped vibrating during incoming calls when running in the background. This issue was caused by the omission of ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL in the getCurrentForegroundType() method for devices running Android Q (API level 29) and above.

By including FOREGROUND_SERVICE_TYPE_PHONE_CALL in the foreground service types, we inform the system that our service handles phone calls. This allows the system to grant necessary capabilities, such as vibration and proper notification handling, when the app is in the background.


dkaraush:
This crashes live streams.

mehebub648:
Okay, but can you find any workaround?

PR #1834, opened 2024-11-30
ChatListCell redrawing fix

maksimillano:
In the **ThemeActivity** in `ChatListCell.ListView#onDraw` method, the color of the button is updated by the method `RadioButton#setColor`, this leads to updating the button itself due to calling `invalidate` method on the button. And the button itself has a overrided `invalidate` method that will lead to the rendering `ChatListCell.ListView` again. Constant unnecessary redrawing occurs. I added a check so that the colors don't update every `onDraw` call.

https://github.com/user-attachments/assets/2309f256-80a0-4e5b-8f98-22de556eba8c


hedayatsayed580-ship-it (commented):
2ac2faa923b80f12ad0ec67d3bb8cbea40d0b0dc

PR #1837, opened 2024-12-22
complete contest android 2024 round 2

mokhtarabadi:
First, I reviewed the code and realized that many parts could be reused.  
That's why I decided to utilize as much of the existing code as possible.  

For the preview, I used the same **photoViewer**, but instead of modifying 20,000 lines of code to add the collage feature, I took a simpler approach: I rendered the collage image or video before it reached the **photoViewer**.  

### Known Issues (if I had more time, I would definitely address them):  
1. **Landscape Mode**:  
   - Touch events for **recordControl** do not work properly.  
   - The collage feature does not function correctly in landscape mode.  

These issues occurred because I considered the landscape mode too late in the process.  

2. **Dual Camera**:  
   - It does not render correctly due to the **bounds** of its parent.

### DIFF
--- a/TMessagesProj/google-services.json
+++ b/TMessagesProj/google-services.json
@@ -68,7 +68,7 @@
       "client_info": {
         "mobilesdk_app_id": "1:760348033671:android:7396e651423888c3f66e22",
         "android_client_info": {
-          "package_name": "org.telegram.messenger.web"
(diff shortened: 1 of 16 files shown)


mokhtarabadi:
sorry i make a mistake :D

PR #1838, opened 2025-01-17
Fix query decoding when opening link

ARTI1208:
If the link contains encoded query parameter e.g. other url with query parameters it is automatically decoded, so the query parameters of inner url will become query parameters of the outer one


Mahhabubhasanhanif:
[](url)

Mahhabubhasanhanif:
1. #1838

Mahhabubhasanhanif:
`

`

Mahhabubhasanhanif:
♥️qqq

Arash63Dehghan (approved):

PR #1846, opened 2025-03-18
Fix bounds check in name_parse function

yannaingtun:
**Description**
Added a bounds check before memcpy operation in name_parse() function to improve security. This ensures that label_len bytes can be safely read from the packet buffer before copying memory. The fix prevents potential out-of-bounds reads when processing malformed DNS packets.

**The fix adds:**
if (j + label_len > length) return -1;

**References:**
Original fix: libevent/libevent@96f64a0
CVE-2016-10195: https://nvd.nist.gov/vuln/detail/CVE-2016-10195


Arash63Dehghan (approved):

yannaingtun:
Hi @Arash63Dehghan, thanks for reviewing and approving the PR! I just wanted to check if you're planning to merge it. Please let me know if you have any concerns.

PR #1881, opened 2025-07-21
Improve Proxy Rotation: Connect to First Available Proxy Without Waiting for All Pings

Kingvornex:
📝 Pull Request Description:

This PR improves the proxy rotation mechanism in Telegram by reducing delay when switching proxies.

✨ Changes:

Immediately connects to the first available proxy once it successfully responds to a ping.

Avoids waiting for all proxy ping checks to complete before switching.

Uses a hasSwitched flag to ensure only one switch happens per rotation check.

Retains switchToAvailable() as a fallback if no proxy responds.

📈 Benefits:

Much faster proxy switching during network instability or failover.

Better experience in regions with partial blocking or when using unreliable proxies.

Reduces unnecessary wait times by prioritizing responsiveness over completeness.


yashikada:
It looks like you used ChatGPT to generate the code.
The comments have the usual emojis that ChatGPT uses when generating the code.

BOplaid:
😐

PR #1882, opened 2025-07-21
fixed

tg2sipfork:
fixed


yashikada:
Have you tried compiling your PR? 
I don't think so. 
In C++, comments aren't added with the # character, but with // or /* */.

PR #1895, opened 2025-09-10
label Profile music player with track title and artist

mehranlatifi83:
Problem
On the Profile screen, the music player that users add to their profiles had no label for screen readers, so nothing meaningful was announced.
Change
In ProfileActivity (profile adapter, VIEW_TYPE_MUSIC case), after setMusicDocument(...) a two-line contentDescription is set:
• 
Line 1: existing Music string
• 
Line 2: existing AccDescrMusicInfo(artist, title) format (same as chats), with fallbacks to AudioUnknownArtist and AudioUnknownTitle when empty.
Result
Screen readers now announce “Music” followed by “{artist}, {title}”. If no saved music is present, the label is cleared to avoid incorrect announcements.
Scope
Only ProfileActivity.java; no new strings; no visual/behavioral changes for sighted users.
Testing

https://github.com/user-attachments/assets/1ec6162f-661e-476d-b8fa-c949b452b051

Built and verified with TalkBack; a profile with saved music is announced with the expected title and artist.


phanvanduoc:
> Problem On the Profile screen, the music player that users add to their profiles had no label for screen readers, so nothing meaningful was announced. Change In ProfileActivity (profile adapter, VIEW_TYPE_MUSIC case), after setMusicDocument(...) a two-line contentDescription is set: • Line 1: existing Music string • Line 2: existing AccDescrMusicInfo(artist, title) format (same as chats), with fallbacks to AudioUnknownArtist and AudioUnknownTitle when empty. Result Screen readers now announce “Music” followed by “{artist}, {title}”. If no saved music is present, the label is cleared to avoid incorrect announcements. Scope Only ProfileActivity.java; no new strings; no visual/behavioral changes for sighted users. Testing
> 
>  demo_for_added_label_to_music_of_profile.mp4 
> Built and verified with TalkBack; a profile with saved music is announced with the expected title and artist.

PR #1896, opened 2025-10-01
Fix audio focus

arsLan4k1390:
Fixed audio focus so that if you are listening to music and start playing a voice message, the music pauses and automatically resumes after the message ends.


scutuatua-crypto (approved):

mio-19:
I listened to music via another app. With this patch the music was not paused when I played a voice message

PR #1902, opened 2025-10-23
Fixes: Attempt fixes for android auto notification

null-nick:


Amnat-auyana1998 (approved):

null-nick:
fixed in 12.7.3 (6751)

PR #1928, opened 2026-02-02
Enable "System Default" Auto Night Mode for all Android versions

Ephemera42:
Since [night mode](https://developer.android.com/reference/android/app/UiModeManager#setNightMode(int)) support was introduced as early as API 8, the "System Default" option can function correctly on older Android versions and should not be restricted to Android 10+ devices.

https://github.com/user-attachments/assets/042532bc-d54a-4096-b83c-1a9d3d0b121b


nabahindah3-cyber:
- - [ ] https://github.com/users/DrKLO/achievements/arctic-code-vault-contributor

PR #1947, opened 2026-03-23
TMA: repeated camera/microphone permission dialog on every WebView PermissionRequest in the same session

T-Damer:
In Telegram Mini Apps on Android, camera/microphone access inside the WebView may trigger the Telegram permission dialog repeatedly within the same Mini App session.

This is especially visible when a Mini App uses `getUserMedia()` more than once, for example:
- starting camera access again after a previous attempt
- switching `facingMode`
- re-requesting audio/video capture from the same page

### Expected behavior

If the user has already approved camera/microphone access for the current Mini App WebView session, and Android runtime permissions are still granted, repeated `PermissionRequest`s should be granted without showing the Telegram permission dialog again.

### Actual behavior

Telegram shows the WebView permission dialog again for subsequent `PermissionRequest`s, even when:
- the same Mini App session is still active
- the same WebView is used
- Android `CAMERA` / `RECORD_AUDIO` permissions are already granted

This makes camera UX unstable for Mini Apps and may interrupt flows like video recording or camera switching.

### Scope

This is about **Telegram Mini Apps / WebView**, not the native attach menu camera flow.

### Reproduction

1. Open a Telegram Mini App on Android.
2. Let it request camera or camera+microphone access.
3. Approve the Telegram dialog and the Android permission request.
4. Trigger another WebView media permission request in the same page/session
   - for example by switching camera or re-requesting media capture.
5. Observe that Telegram shows its permission dialog again.

### Notes

The issue appears to be in WebView `onPermissionRequest(...)` handling in the Android client. Issue is not present on iOS client

A safe fix is to auto-grant repeated WebView media requests **only inside the same Mini App session**, when:
- the WebView session has already been approved by the user
- and Android runtime permissions are still granted

This should not bypass permission prompts for a fresh page load or a new Mini App session.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/ui/web/BotWebViewContainer.java
+++ b/TMessagesProj/src/main/java/org/telegram/ui/web/BotWebViewContainer.java
@@ -682,6 +682,36 @@ public abstract class BotWebViewContainer extends FrameLayout implements Notific
         }
     }
 
+    @RequiresApi(api = Build.VERSION_CODES.LOLLIPOP)
+    private String[] getRequestedAndroidPermissionsForWebResources(String[] resources) {
+        ArrayList<String> permissions = new ArrayList<>();
+        for (String resource : resources) {
+            if (PermissionRequest.RESOURCE_AUDIO_CAPTURE.equals(resource)) {
+                if (!permissions.contains(Manifest.permission.RECORD_AUDIO)) {
+                    permissions.add(Manifest.permission.RECORD_AUDIO);
+                }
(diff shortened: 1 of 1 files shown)


xmagor:
Did anyone notice this? It creates quite a bit of friction that he keeps asking for permission for the same thing within the same session.

thatkit:
The issue still persits. Devs, please take actions on the PR above

abtopeza-beep (approved):

umidjon-2231:
any updates?

dehghanpur10:
I’m seeing the same issue. Could you please fix it? Thanks!

PR #1959, opened 2026-05-22
Fix Telegram proxy connectivity over IPv6

1223421:
Fix several issues that prevented Telegram from working correctly through IPv6 proxies.


MaxS0niX:
A fix to
https://bugs.telegram.org/c/16335
https://bugs.telegram.org/c/19197
https://bugs.telegram.org/c/29697
https://bugs.telegram.org/c/60319

PR #1971, opened 2026-06-19
Setup initial configs

isaac-marlow:


isaac-marlow:
Sorry, mistake by my side

PR #2008, opened 2026-08-03
Migrate Android Auto implementation to androidx.car.app:app:1.7.0

rainman74:
## What
Aligns the Android Auto code path in `DrKLO/Telegram` with the `androidx.car.app:app:1.7.0` contract, and adds the `MusicPlayerService -> TelegramMediaSession` publish hooks so that the CarApp / MediaBrowserService surface is in sync with the phone player.

## Why
The 1.7.x line of androidx.car.app tightened the contract that CarAppService implementations have to satisfy:
- `ConversationItem.validateSender()` now requires every `Person` passed in to have a non-null key. A missing key throws at runtime.
- The new contract needs `androidx.media:1.7.0` (for the `android.support.v4.media.*` compat shims) and `androidx.lifecycle:lifecycle-runtime:2.7.0` (for `DefaultLifecycleObserver` used by `HomeScreen`).
- `minCarApiLevel` has to be bumped from 5 to 7.

The phone player's `MusicPlayerService` keeps its own `MediaSessionCompat`, but the CarApp / MediaBrowserService surface that Android Auto uses is backed by `TelegramMediaSession`. Without explicitly publishing playback state and metadata to `TelegramMediaSession`, the Android Auto side has no view of what the phone player is doing. The same goes for `STATE_STOPPED` on service teardown.

## How
**6 files changed (+23/-10 vs `master`):**

### Build
- `TMessagesProj/build.gradle` - bump `androidx.car.app:app:1.4.0 -> 1.7.0`; drop `app-projected:1.4.0`; add `androidx.media:1.7.0` and `androidx.lifecycle:lifecycle-runtime:2.7.0`

### Manifest / config
- `TMessagesProj/src/main/AndroidManifest.xml` - uncomment the CarApp `<meta-data>` + `<service>` block, bump `minCarApiLevel` 5 -> 7
- `TMessagesProj/src/main/res/xml/automotive_app_desc.xml` - add `<uses name="media"/>`

### Code
- `TMessagesProj/src/main/java/org/telegram/messenger/car/HomeScreen.java` - the `senderBuilder.setName("")` empty-fallback branch in `buildCarMessage` now also calls `setKey("d" + dialogId)`, so every `Person` handed to a `ConversationItem` has a key.
- `TMessagesProj/src/main/java/org/telegram/messenger/MusicPlayerService.java` - after every `mediaSession.setPlaybackState()` and `mediaSession.setMetadata()`, also `publishPlaybackState()` / `publishMetadata()` on `TelegramMediaSession`; on `onDestroy` publish `STATE_STOPPED` when nothing is playing.
- `TMessagesProj/src/main/java/org/telegram/messenger/TelegramMediaSession.java` - `publishPlaybackState` now merges `getAvailableActions()` into the state so Android Auto renders the full action set.

## Notes
- The `androidx.media:1.7.0` artifact keeps its classes under the legacy `android.support.v4.media.*` package for binary compatibility - no import migration needed.
- The `MusicPlayerService` / `TelegramMediaSession` changes are applied as targeted hunks on top of upstream `master`, not as a verbatim file copy, because upstream `master` carries additional MIUI-specific lock-screen logic in `supportLockScreenControls` and an `accountLogin` handler in `didReceivedNotification` that the NagramX-derived forks do not have. Only the contract-alignment hunks are taken; the rest of the upstream file is preserved as-is.
- This PR is intentionally limited in scope: it does not touch the upstream `accountLogin` handler in `MusicPlayerService` (which is the NagramX-specific quirk that is being removed there). If upstream later wants to follow the same removal, that should be a separate change with its own discussion.
- Final diff against master: 6 files, 23 insertions(+), 10 deletions(-).

### DIFF
--- a/TMessagesProj/build.gradle
+++ b/TMessagesProj/build.gradle
@@ -27,9 +27,10 @@ dependencies {
 
     implementation 'com.google.android.gms:play-services-cast-framework:21.4.0'
     implementation "androidx.mediarouter:mediarouter:1.7.0"
+    implementation "androidx.media:media:1.7.0"
+    implementation "androidx.lifecycle:lifecycle-runtime:2.7.0"
 
-    implementation 'androidx.car.app:app:1.4.0'
-    implementation 'androidx.car.app:app-projected:1.4.0'
+    implementation 'androidx.car.app:app:1.7.0'
     implementation 'org.nanohttpd:nanohttpd:2.3.1'
 
     compileOnly 'org.checkerframework:checker-qual:2.5.2'
--- a/TMessagesProj/src/main/AndroidManifest.xml
+++ b/TMessagesProj/src/main/AndroidManifest.xml
@@ -672,7 +672,6 @@
 
         <meta-data android:name="com.google.android.gms.wallet.api.enabled" android:value="true" />
 
-        <!--
         <meta-data android:name="com.google.android.gms.car.notification.SmallIcon" android:resource="@drawable/ic_player" />
         <meta-data android:name="com.google.android.gms.car.application" android:resource="@xml/automotive_app_desc" />
 
@@ -688,8 +687,7 @@
 
         <meta-data
             android:name="androidx.car.app.minCarApiLevel"
-            android:value="5" />
-        -->
+            android:value="7" />
 
         <meta-data android:name="com.google.android.gms.vision.DEPENDENCIES" android:value="face,barcode" />
 
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MusicPlayerService.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MusicPlayerService.java
@@ -526,7 +526,9 @@ public class MusicPlayerService extends Service implements NotificationCenter.No
                 }
             }
 
-            mediaSession.setPlaybackState(playbackState.build());
+            PlaybackStateCompat currentPlaybackState = playbackState.build();
+            mediaSession.setPlaybackState(currentPlaybackState);
+            TelegramMediaSession.getInstance(this).publishPlaybackState(currentPlaybackState);
             updateRepeatMode();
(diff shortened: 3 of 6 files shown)


rainman74:
Update: corrected the import migration in commit `43c3a52e`.

The original commit `32ca438` bumped `androidx.media:media:1.7.0` and migrated the imports to `androidx.media.*`, but that artifact does **not** publish its `MediaSessionCompat` / `MediaDescriptionCompat` / `MediaMetadataCompat` / `RatingCompat` / `MediaControllerCompat` / `PlaybackStateCompat` classes under `androidx.media.*` — they stay under the legacy `android.support.v4.media.*` package for binary-compatibility reasons. Only the newer `AudioAttributesCompat` / `AudioFocusRequestCompat` / `VolumeProviderCompat` family is in the `androidx.media.*` package.

Verified by downloading the actual `androidx.media:media:1.7.0` AAR from `dl.google.com` and listing the `classes.jar` entries:

- `android.support.v4.media.*` — **226 classes** (including `MediaSessionCompat`, `MediaDescriptionCompat`, `MediaMetadataCompat`, `RatingCompat`, `MediaBrowserCompat`, `PlaybackStateCompat`, `MediaControllerCompat`)
- `androidx.media.*` — **94 classes** (only `AudioAttributesCompat`, `AudioAttributesImpl*`, `AudioFocusRequestCompat`, `VolumeProviderCompat` family)

So the original import migration would not have compiled.

**Fix:** force-pushed commit `43c3a52e` reverts the seven import files back to `android.support.v4.media.*`. The transitive `androidx.media:media:1.7.0` dependency stays (it is required for the dependency declaration of the manifest + lifecycle observer); we just don't migrate the imports.

Final diff vs `dev`: 4 files, +7/-7 (the seven import-migration files cancel out vs. master).

PR #2013, opened 2026-08-13
Fix slow media viewer opening and jumping counter in chats

dunxuan:
Opening a photo or video in a chat was slow for older media: the viewer searched for the target message page by page starting from the newest media, so the further back the media was, the longer it took to load the list, the counter and swiping.

The media counter (x/y) could also jump (e.g. 3/43 -> 3/42) because hidden sensitive media (isHiddenSensitive) was filtered out of the media list while the server total still counted it.

## Changes
- **Load media from around the target message** (max_id = target id + 1) and expand the list in both directions, instead of searching page by page from the newest media
- **Build the initial list from messages already in memory** in ChatActivity, so the viewer can be swiped immediately without waiting for the server
- **Keep hidden sensitive media in the list** (rendered as a placeholder, not loading the actual image), matching the iOS behaviour, so position/count stay consistent
- **Only show the counter once an exact value is known** (either an edge is reached or the local cache covers all media), so it never shows wrong or jumping values
- **Add a local DB query** (media_v4 counts around the target id, exposed via a new mediaPositionCountsDidLoad notification) to resolve the exact position quickly when the local cache covers all media

## Scope
Only the regular chat flow is affected (non-encrypted, non-thread, non-filtered dialogs); other entry points keep the original behaviour.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MessagesStorage.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MessagesStorage.java
@@ -15144,6 +15144,36 @@ public class MessagesStorage extends BaseController {
         }
     }
 
+    public void getMediaPositionCounts(long dialogId, int type, int mid, int classGuid) {
+        storageQueue.postRunnable(() -> {
+            int before = 0;
+            int after = 0;
+            try {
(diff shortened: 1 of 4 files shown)


dunxuan:
Closing this PR - the in-memory initial list approach causes issues when opening media from comments (it collects media across different parent messages). Will take a different approach.

PR #2021, opened 2026-08-24
Say what the update bar is and how far its download has got

mehranlatifi83:
## Summary

A new version announces itself with a dialog that reads well enough, and pressing update on it leaves a bar at the bottom of the main screen. That bar is the only way to follow the download, to stop it, and to install what came down, and a screen reader has nothing to say about it: the text it shows is drawn rather than laid out, the size beside it is drawn by something that is not a view at all, and so is the icon that carries its state.

Have the bar say what it draws. It reports as the button it is, it says the text it shows along with the size of what is to be downloaded, and the label of its click action says what pressing it does at that moment, which is downloading, stopping, or installing, all of which its icon is alone in telling everyone else. The text it holds is left out of what is read, so the bar is one stop rather than two.

Both builds that carry an update bar of their own are covered.

## Checking it

With TalkBack on, take an in-app update and press update on the dialog, which leaves a bar at the bottom of the main screen.

Before, that bar said nothing at all: not what it was, not the size of what was coming down, not how far it had got. Now it reports as the button it is, says its text and the size beside it, and its action says what pressing it does at that moment, which is downloading, stopping or installing.

### DIFF
--- a/TMessagesProj_AppHockeyApp/src/main/java/org/telegram/ui/Components/UpdateLayout.java
+++ b/TMessagesProj_AppHockeyApp/src/main/java/org/telegram/ui/Components/UpdateLayout.java
@@ -9,6 +9,7 @@ import android.graphics.Canvas;
 import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.accessibility.AccessibilityNodeInfo;
 import android.widget.FrameLayout;
 
 import org.telegram.messenger.AndroidUtilities;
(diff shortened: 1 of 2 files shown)


pfeux:
@DrKLO  The build currently fails during Gradle sync because the jlatexmath module is not included in the root settings.gradle.

Please add:

```gradle
include ':jlatexmath'
project(':jlatexmath').projectDir = file('TMessagesProj/lib/jlatexmath/jlatexmath')

mehranlatifi83:
Hi @pfeux — this is about the repository's own `settings.gradle`, not about the change in this pull request, and it has since been fixed upstream: `12.10.1` includes `:jlatexmath` and sets its `projectDir` exactly as you wrote. A fresh checkout of master syncs without that edit now. Note that `third_party` moved to submodules in the same release, so `git submodule update --init --recursive` is needed before building. This branch has just been rebased onto `12.10.1`.

PR #2022, opened 2026-08-24
Fix missing :jlatexmath module include in settings.gradle

null-nick:


hedayatsayed580-ship-it:
null-nick:fix/jlatexmath-settings-include

PR #2028, opened 2026-08-25
Fix slow media viewer opening in chats

dunxuan:
### Problem
Opening a photo or video directly in a chat was slow for older media: the viewer located the target message by paging backwards from the newest media, so the further back the media was, the longer it took until the list, the counter and swiping became usable. In chats with lots of media the counter often never appeared at all. Opening media from a comment thread (reply chat) could even jump to media of another parent message.

### Changes
- **Locate media around the target message**: the initial request uses max_id = target id + 1 so the target is inside the first page, then the list is expanded in both directions (older/newer) in 200-item pages, instead of paging from the newest media
- **Build the initial list from in-memory messages**: ChatActivity collects all media of the same type already loaded in the chat (including comment threads) and passes them to the viewer, so it can be swiped immediately without waiting for the server
- **Keep hidden sensitive media in the list** (rendered as a placeholder instead of loading the image) so positions and counts stay consistent, matching iOS behaviour
- **Only show the counter once it is exact**: either an edge is reached, or a local DB query (media_v4, new mediaPositionCountsDidLoad notification) resolves the exact position when the local cache covers all media; otherwise the counter stays hidden instead of showing wrong values
- Comment threads use the in-memory list but skip the DB counter (media_v4 has no topic dimension), falling back to exact counts once an edge is reached

### Scope
Only the regular chat flow (non-encrypted, non-filtered, non-scheduled, non-sponsored media) is affected; other entry points keep the original behaviour.

### DIFF
--- a/TMessagesProj/src/main/java/org/telegram/messenger/MessagesStorage.java
+++ b/TMessagesProj/src/main/java/org/telegram/messenger/MessagesStorage.java
@@ -15144,6 +15144,36 @@ public class MessagesStorage extends BaseController {
         }
     }
 
+    public void getMediaPositionCounts(long dialogId, int type, int mid, int classGuid) {
+        storageQueue.postRunnable(() -> {
+            int before = 0;
+            int after = 0;
+            try {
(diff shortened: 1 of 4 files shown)


dunxuan:
Closing - found a counter issue in comment threads (shows 0/xx before 1/xx). Will investigate and reopen with a fix.

PR #2049, opened 2026-09-03
Loop emoji statuses forever in group/voice call participant list

anriltine:
## Summary
- Group/voice call participant emoji statuses currently use `CACHE_TYPE_ALERT_EMOJI_STATUS`, which caps animation at **2 loops**, so custom video emoji freeze mid-call while gift overlays keep animating.
- Switch `GroupCallUserCell` left/right status drawables to `CACHE_TYPE_EMOJI_CALL`, which already configures infinite `autoRepeatCount` for call UI.

## Test plan
- [ ] Join a group voice chat with a Premium emoji status set (video custom emoji preferred)
- [ ] Confirm the status next to the name keeps looping for the whole call
- [ ] Confirm verified / premium-star fallbacks still render
- [ ] Confirm chat-list emoji statuses still stop after ~2 loops (unchanged)

Made with [Cursor](https://cursor.com)


anriltine:
Cross-platform notes:
- iOS equivalent fix: https://github.com/TelegramMessenger/Telegram-iOS/pull/2319
- iOS issue: https://github.com/TelegramMessenger/Telegram-iOS/issues/2320
- Desktop discussion: https://github.com/telegramdesktop/tdesktop/issues/31236

Root cause is the same intentional 2-loop cap for emoji statuses; call UI already had an infinite path on Android (`CACHE_TYPE_EMOJI_CALL`) that this PR wires up for the participant list.

